PR #36, opened 2014-08-10
App package

prathamesh920:
created a setup file to install the exam app using the following commands:
pip install django-exam-0.1.tar.gz
pip install git+https://github.com/prathamesh920/online_test.git#egg=django_exam-0.1


prabhuramachandran:
Again, please do not check-in the django-exam-0.1.tar.gz into the repo.

prabhuramachandran:
The PR #37 is similar to this, which of these is to be merged?

PR #37, opened 2014-08-11
Redirect result view

prathamesh920:
Redirect to user dashboard after the exam is over without logging out the user.
After the exam, user can go back but cannot reattempt the same exam again. 
Fixed minor issues in css.


testapp/setup.py
```diff
@@ -0,0 +1,39 @@
...
+
+README = open(os.path.join(os.path.dirname(__file__), 'README.rst')).read()
+
+# allow setup.py to be run from any path
+os.chdir(os.path.normpath(os.path.join(os.path.abspath(__file__), os.pardir)))
+
+setup(
+    name='django-exam',
+    author='Prabhu Ramachandran',
+    author_email='prabhu.ramachandran@gmail.com',
```
prabhuramachandran: This should really be changed to Python team or something like that.

testapp/setup.py
```diff
@@ -0,0 +1,39 @@
...
+    version='0.1',
+    packages=['exam'],
+    include_package_data=True,
+    license='BSD License',
+    description='A django app to conduct online test.',
```
prabhuramachandran: "conduct online test." -> "conduct online tests."

testapp/LICENSE
```diff
@@ -0,0 +1,3 @@
+This is distributed under the terms of the BSD license. 
+
+Copyright (c) 2011 Prabhu Ramachandran and FOSSEE (fossee.in)
```
prabhuramachandran: Change this to 2011-2015

prabhuramachandran:
You seem to have checked in a tarball of the sources into testapp/dist, this is not good.  The tarball should be hosted either on github's downloads page or on pypi.

prabhuramachandran:
Please update this PR as per comments.

PR #38, opened 2015-02-20
Code server redesign mymaster

ankitjavalkar:


ankitjavalkar:
Currently need to add views and templates hence closing for now.

PR #40, opened 2015-03-04
Redirect result view

prathamesh920:
Redirect to user dashboard after the exam is over without logging out the user.
After the exam, user can go back but cannot reattempt the same exam again.
Fixed minor issues in css.
README modified.
Fixed a logical issue in random questions for question paper generation.


testapp/README.rst
```diff
@@ -10,7 +10,10 @@ conduct online test and monitor the test.
 Quick start
 ------------
 
-1. Add "testapp.exam", "taggit" and "taggit_autocomplete_modified" apps 
+1. In yourterminal run the following command::
```
prabhuramachandran: "your terminal"

testapp/README.rst
```diff
@@ -30,11 +33,11 @@ Quick start
...
 
 
-4. Run 'python manage.py syncdb' to create models for the new installed apps.
+5. Run 'python manage.py syncdb' to create models for the new installed apps.
 
-5. Start the development server and visit http://localhost:8000/exam/
+6. Start the development server and visit http://localhost:8000/exam/
 
-6. In exam app run code_sever command  as superuser as follows::
+7. In exam app run code_sever command  as superuser as follows::
```
prabhuramachandran: What is this "in exam app" mean?  Be specific.

testapp/README.rst
```diff
@@ -30,11 +33,11 @@ Quick start
...
 
        $ code_server
```
prabhuramachandran: sudo code_server?

prathamesh920: Sir, if the app is installed in a python virtual environment then sudo code_server command will look for the command in the system bin folder and therefore will not work. So the user has to activate the vitual environment as superuser and then just run code_server, it will work.
sudo code_server will be fine, if there is no virtual environment.

prabhuramachandran:
Merging for now but the javascript code smells badly.

PR #46, opened 2015-04-28
Update README

ankitjavalkar:


README.md
```diff
@@ -1,93 +1,113 @@
...
+ * Since it runs on your Python, you could technically test any Python based library.
+ * Scales to over 500+ simultaneous users.
+ * Distributed under the BSD license.
+
+Quick Start for Users
+======================
+
+#### Pre-Requisite
+
+1. [Install pip](https://pip.pypa.io/en/latest/installing.html)
```
prabhuramachandran: Make sure you have pip available and usable might be a better usage.

So what does someone on a Linux machine without root access do?  There should be a section that you can link to for the benefit of those users.

README.md
```diff
@@ -1,93 +1,113 @@
...
-  
+#### Installation
+
+1. Run the install_script.py using
+    ```python install_script.py```
+
+    The script does the following;
+    1. Creates a new virtual environment called online_test
+    1. Installs Django 1.6 within that virtual environment
+    1. Pulls from the online_test repository on [Github](https://github.com/FOSSEE/online_test)
```
prabhuramachandran: Don't you need the online_test repo to get the install_script in the first case?  This is a tad too complex.  If we upload online_test on pypi it should be pip install online_test or something.

Also can't you point pip to the git repo directly and be done with it?  This section needs a lot of work as this is no easy to use quick start.

README.md
```diff
@@ -1,93 +1,113 @@
...
+    1. Creates a two users test moderator and a test examinee
+    1. Loads demo questions
+    1. Loads demo quiz
+
+1. In your terminal run ```launch demo```
```
prabhuramachandran: What is `launch`?  If it is not standard django please do not introduce new commands that only confuse users.

prabhuramachandran:
This is not nearly enough.  It has to be very simple to start with and the current instructions are not.  The accompanying code changes are also not in place so I will not merge this until it actually works as described.  Closing for now.

PR #48, opened 2015-05-13
Re apply docs directory in testapp directory and other changes

ankitjavalkar:
Needs to be merged First - before #49

PR #49, opened 2015-05-27
Feedback changes

ankitjavalkar:
Only to be merged after Pull Request #48

### DIFF
--- a/testapp/exam/models.py
+++ b/testapp/exam/models.py
@@ -1,6 +1,7 @@
 import datetime
 import json
 from random import sample, shuffle
+from itertools import islice, cycle
 from django.db import models
 from django.contrib.auth.models import User
 from taggit_autocomplete_modified.managers import TaggableManagerAutocomplete\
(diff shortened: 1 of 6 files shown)


prabhuramachandran:
Not sure what is going on, as this still muddles up with the PR #48.  Can you either merge master to this branch or rebase to get a clean PR?

testapp/exam/models.py
```diff
@@ -317,9 +322,14 @@ class AnswerPaper(models.Model):
     status = models.CharField(max_length=20, choices=test_status,\
             default='inprogress')
 
+    # def initialise_questions_unanswered(self):
```
prabhuramachandran: No commented code please.

testapp/exam/models.py
```diff
@@ -298,6 +300,9 @@ class AnswerPaper(models.Model):
     # The questions successfully answered (a list of ids separated by '|')
     questions_answered = models.CharField(max_length=128)
 
+    # The unanswered questions (a list of ids separated by '|')
+    questions_unanswered = models.CharField(max_length=128)
```
prabhuramachandran: Why do you need a new model attribute?  All you need is a new method:
`get_unanswered_questions` which does whatever necessary.

PR #56, opened 2015-08-28
Start and Expiry times for Quizzes

ankitjavalkar:

PR #57, opened 2015-09-28
API initial cut - Please merge only after #56 PR

ankitjavalkar:


yaksh/api/yaksh_api.py
```diff
@@ -0,0 +1,150 @@
...
+                )
+
+    def detail(self, pk):
+        return Question.objects.get(id=pk)
+
+    def list(self):
+        return Question.objects.all()
+
+    def delete(self, pk):
+        Question.objects.get(id=pk)
```
prabhuramachandran: Looks wrong.

PR #60, opened 2015-10-08
Add missing evaluator tests

ankitjavalkar:
Had been missed out when moving the directories and files around. Added now.


.travis.yml
```diff
@@ -9,7 +9,7 @@ env:
 
 # command to install dependencies
 install:
-  - "easy_install git+https://github.com/FOSSEE/online_test.git#egg=django_exam-0.1"
+  - "easy_install git+https://github.com/FOSSEE/online_test.git#egg=yaksh-0.1"
```
prabhuramachandran: Use pip

PR #62, opened 2015-10-19
add a rest base api based on Restless python package

ankitjavalkar:
This API is tied down to Django and is provided as a webservice


ankitjavalkar:
Need to work on this further closing for now

PR #63, opened 2015-10-19
API

ankitjavalkar:
This feature can be run from command line, similar to a API.
Not a web service
Is not linked to URLs


ankitjavalkar:
Need to work on this further, closing for now

PR #66, opened 2015-11-23
Allow single user session active at any given time

ankitjavalkar:


yaksh/models.py
```diff
@@ -8,6 +8,12 @@
 
 
 ###############################################################################
+class Visitor(models.Model):
```
prabhuramachandran: Visitor commonly is used for a Visitor pattern, so  I would suggest a different name.

yaksh/middleware/one_session_per_user.py
```diff
@@ -0,0 +1,21 @@
...
+from django.contrib.sessions.models import Session
+
+from yaksh.models import ConcurrentUser
+
+
+class OneSessionPerUserMiddleware(object):
+    def process_request(self, request):
+        if isinstance(request.user, User):
+            current_key = request.session.session_key
+            if hasattr(request.user, 'concurrentuser'):
```
prabhuramachandran: `concurrent_user`?

yaksh/middleware/one_session_per_user.py
```diff
@@ -0,0 +1,21 @@
...
+                active_key = request.user.concurrentuser.session_key
```
prabhuramachandran: `concurrent_user`?

yaksh/middleware/one_session_per_user.py
```diff
@@ -0,0 +1,21 @@
...
+                if active_key != current_key:
+                    Session.objects.filter(session_key=active_key).delete()
+                    request.user.concurrentuser.session_key = current_key
+                    request.user.concurrentuser.save()
+            else:
+                ConcurrentUser.objects.create(
```
prabhuramachandran: Something seems very wrong here.

PR #83, opened 2016-03-08
Forgot password

prathamesh920:
Used django in-built views.
Templates overridden.
User enters email address and submits.
The user receives an email with password reset link(one time link).
The link contains a token generate using the current state of the user
like user password.
The link is verified each time when it is been requested.
So the link will be invalid if the user has already changed the password
using the link.(since the token is generated using current state of the user.)
User resets his password via the link.


prathamesh920:
forgot_password
after_email_submit
reset_password
after_password_reset

PR #87, opened 2016-03-21
Enrollment and profile changes

adityacp:
added edit profile facility for moderators and students
added change password facility for moderators and students
enroll all students
moderator can see user profile in tooltip

### DIFF
--- a/yaksh/forms.py
+++ b/yaksh/forms.py
@@ -291,6 +291,19 @@ class EditProfile(forms.Form):
     position = forms.CharField\
         (max_length=64)
 
+    def save(self, user):
+        user_details = User.objects.get(id=user.id)
+        profile_details = Profile.objects.get(user_id=user.id)
+        user_details.first_name = self.cleaned_data['first_name']
+        user_details.last_name = self.cleaned_data['last_name']
+        profile_details.department = self.cleaned_data['department']
+        profile_details.institute = self.cleaned_data['institute']
+        profile_details.roll_number = self.cleaned_data['roll_number']
+        profile_details.position = self.cleaned_data['position']
+        profile_details.save()
+        user_details.save()
+
+
 class PasswordResetForm(forms.Form):
     """ change password form for students and moderators """
 
--- a/yaksh/forms.py
+++ b/yaksh/forms.py
@@ -276,3 +276,27 @@ class CourseForm(forms.ModelForm):
     class Meta:
         model = Course
         fields = ['name', 'active', 'enrollment']
+
+class EditProfile(forms.Form):
+    """ edit profile form for students and moderators """
+
+    first_name = forms.CharField(max_length=30)
+    last_name = forms.CharField(max_length=30)
+    institute = forms.CharField\
+                (max_length=128)
+    department = forms.CharField\
+                (max_length=64)
+    roll_number = forms.CharField\
+                (max_length=30)
+    position = forms.CharField\
+        (max_length=64)
+
+class PasswordResetForm(forms.Form):
+    """ change password form for students and moderators """
+
+    new_password = forms.CharField(widget=forms.PasswordInput(attrs={
+                               'class': 'form-control',
+                               'placeholder': 'New Password'}), label='')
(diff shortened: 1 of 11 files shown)


yaksh/forms.py
```diff
@@ -276,3 +276,40 @@ class CourseForm(forms.ModelForm):
     class Meta:
         model = Course
         fields = ['name', 'active', 'enrollment']
+
+class EditProfile(forms.Form):
```
prathamesh920: you can use Profile model forms, this will make saving and editing easy for you.

yaksh/forms.py
```diff
@@ -276,3 +276,40 @@ class CourseForm(forms.ModelForm):
...
+        user_details.last_name = self.cleaned_data['last_name']
+        profile_details.department = self.cleaned_data['department']
+        profile_details.institute = self.cleaned_data['institute']
+        profile_details.roll_number = self.cleaned_data['roll_number']
+        profile_details.position = self.cleaned_data['position']
+        profile_details.save()
+        user_details.save()
+
+
+class PasswordResetForm(forms.Form):
```
prathamesh920: once check django built-in password reset. Then there will not be need to wite a view for this.

yaksh/views.py
```diff
@@ -1229,20 +1229,33 @@ def courses(request):
...
+def enroll(request, course_id, user_id=None, was_rejected=False):
     user = request.user
+    ci = RequestContext(request)
     if not is_moderator(user):
         raise Http404('You are not allowed to view this page')
     course = get_object_or_404(Course, creator=user, pk=course_id)
-    user = get_object_or_404(User, pk=user_id)
-    course.enroll(was_rejected, user)
+    if request.method == 'POST':
+        data = request.POST.getlist('check')
```
prathamesh920: data is a list of user ids so you can give a better name to it.

yaksh/views.py
```diff
@@ -1229,20 +1229,33 @@ def courses(request):
...
+        if data is None:
+            return my_render_to_response('yaksh/course_detail.html', {'course': course},
+                                        context_instance=ci)
+        else:
+            for i in data:
```
prathamesh920: again "i" variable name will change

yaksh/views.py
```diff
@@ -1229,20 +1229,33 @@ def courses(request):
...
+                user = get_object_or_404(User, pk=i)
```
prathamesh920: the user object has to be present since we are only passing the ids through the context.
instead of for loop and checking object exists or not better to use id__in as the filter.

yaksh/views.py
```diff
@@ -1229,20 +1229,33 @@ def courses(request):
...
+                course.enroll(was_rejected, user)
+    else:
+        user = get_object_or_404(User, pk=user_id)
+        course.enroll(was_rejected, user)
```
prathamesh920: note you can pass list of user to the enroll function

yaksh/views.py
```diff
@@ -1698,3 +1711,95 @@ def design_questionpaper(request):
...
+    """ view moderators and users profile """
+
+    context = {}
+    user = request.user
+    ci = RequestContext(request)
+    if not user.is_authenticated():
+        raise Http404('You are not allowed to view this page!')
+    else:
+        profile_details = Profile.objects.get(user_id=user.id)
+        user_details = User.objects.get(id=user.id)
```
prathamesh920: you already have user object with you, so no need to get it.
Also, user.profile can give profile or vice versa, so need to fetch the objects.

yaksh/views.py
```diff
@@ -1698,3 +1711,95 @@ def design_questionpaper(request):
...
+def edit_profile(request):
+    """ edit profile details facility for moderator and students """
+
+    context = {}
+    user = request.user
+    ci = RequestContext(request)
+    data = {}
+    if not user.is_authenticated():
+        raise Http404('You are not allowed to view this page!')
+    profile_details = Profile.objects.get(user_id=user.id)
```
prathamesh920: again u already have user object from request.user

PR #92, opened 2016-04-12
Add teachers to a course 

adityacp:
moderators can search for teachers for a particular course
moderators can add and remove a teacher
teachers can enroll and reject students for the allotted course
teachers can monitor quiz status
teachers can grade user


yaksh/views.py
```diff
@@ -605,7 +606,7 @@ def course_detail(request, course_id):
     user = request.user
     if not is_moderator(user):
         raise Http404('You are not allowed to view this page')
-    course = get_object_or_404(Course, creator=user, pk=course_id)
+    course = Course.objects.filter(Q(creator=user)|Q(teachers=user), pk=course_id)[0]
```
ankitjavalkar: Is there a better way to express this? This will fail in case there is a user who is neither a Creator nor a Teacher related to that course.(indexing an empty list will throw IndexError)

yaksh/models.py
```diff
@@ -116,6 +118,9 @@ def activate(self):
     def deactivate(self):
         self.active = False
 
+    def add_teachers(self, *users):
+        self.teachers.add(*users)
```
ankitjavalkar: Please just make sure that the users that are being added have moderator permissions
(Maybe an IF condition that checks this and returns appropriate value)

yaksh/views.py
```diff
@@ -813,7 +816,8 @@ def download_csv(request, questionpaper_id):
     if not is_moderator(user):
         raise Http404('You are not allowed to view this page!')
     quiz = Quiz.objects.get(questionpaper=questionpaper_id)
-    if quiz.course.creator != user:
+    course = Course.objects.filter(teachers=user, id=quiz.course.id)
+    if quiz.course.creator != user and not course:
```
ankitjavalkar: Another way to do this: if user != quiz.course.creator and user not in quiz.course.teachers

yaksh/views.py
```diff
@@ -954,3 +958,79 @@ def design_questionpaper(request):
...
+    """ add teachers to the course also add students to
+        moderator group if a student is allotted to course """
+
+    user = request.user
+    ci = RequestContext(request)
+    context = {}
+    course = get_object_or_404(Course, creator=user, pk=course_id)
+    context['course'] = course
+    if not user.is_authenticated() or not is_moderator(user):
+        raise Http404('You are not allowed to view this page!')
```
ankitjavalkar: This needs to be checked first, right after obtaining the ```user``` (that is it should be at line 1000)

yaksh/views.py
```diff
@@ -954,3 +958,79 @@ def design_questionpaper(request):
...
+    else:
+        return my_render_to_response('yaksh/addteacher.html', context,
+                                    context_instance=ci)
+
+@login_required
+def view_courses(request):
+    """  show courses allotted to a user """
+
+    user = request.user
+    ci = RequestContext(request)
```
ankitjavalkar: please check if user is a moderator;
```
if not user.is_authenticated() or not is_moderator(user):
```

yaksh/views.py
```diff
@@ -954,3 +958,79 @@ def design_questionpaper(request):
...
+@login_required
+def search_teacher(request, course_id):
+    """ search teachers for the course """
+
+    user = request.user
+    ci = RequestContext(request)
+    context = {}
+    course = get_object_or_404(Course, creator=user, pk=course_id)
+    context['course'] = course
+    if not user.is_authenticated() or not is_moderator(user):
```
ankitjavalkar: 1. This needs to be checked first, right after obtaining the user (that is it should be at line 967)
1. You can now remove ```user.is_authenticated()``` since ```@login_required``` already checks if user is authenticated

yaksh/views.py
```diff
@@ -954,3 +958,79 @@ def design_questionpaper(request):
...
+    if request.method == 'POST':
+        u_name = request.POST.get('uname')
+        if len(u_name) == 0:
+            context['success'] = False
+            return my_render_to_response('yaksh/addteacher.html', context,
+                                        context_instance=ci)
+        else:
+            teachers = User.objects.filter(Q(username__icontains=u_name)|
+                Q(first_name__icontains=u_name)|Q(last_name__icontains=u_name)|
+                Q(email__icontains=u_name)).exclude(is_superuser=1)
```
ankitjavalkar: is it possible to exclude the current user here itself instead of doing it in the template?

yaksh/views.py
adityacp: function "course_detail" will be called only when there is a course either as a teacher or moderator. so list indexing will not throw error.

yaksh/models.py
```diff
@@ -23,6 +23,14 @@ class Profile(models.Model):
     department = models.CharField(max_length=64)
     position = models.CharField(max_length=64)
 
+    def is_moderator(self, user):
+        return True if user.groups.filter(name='moderator').exists() else False
```
prabhuramachandran: cant this be just `return user.groups.filter(name='moderator').exists()`

yaksh/models.py
```diff
@@ -23,6 +23,14 @@ class Profile(models.Model):
...
+
+    def _add_to_group(self, user):
+        group = Group.objects.get(name="moderator")
+        user.groups.add(group)
+        user.profile.position = "Faculty"
```
prathamesh920: No need to change the position in the code. The user can edit his profile.

yaksh/templates/yaksh/addteacher.html
```diff
@@ -0,0 +1,83 @@
...
+			<th>First Name</th>
+			<th>Last Name</th>
+			<th>Email</th>
+			<th>Institute</th>
+			<th>Department</th>
+            <th>Position</th>
+			{% for teacher in teachers %}
+				<tr>
+				<td><input type="checkbox" name="check" value="{{ teacher.id }}"></td>
+				<td>{{ teacher.first_name }}</td>
```
prathamesh920: Can display username as well because it is search by username, firstname. Will be helpful.

yaksh/templates/yaksh/viewcourse.html
```diff
@@ -0,0 +1,51 @@
...
+                    {% else %}
+                        <p><b>No quiz </b>
+                    {% endif %}
+                </div>
+            </div>
+        </div>
+    </div>
+    <br><br> 
+{% endfor %}
+<button class="btn primary" type="button" onClick='location.replace("{{URL_ROOT}}/exam/manage/add_course");'>Add New Course</button>
```
prathamesh920: Add quiz or course buttons not needed in the view allotted courses. Anyway we display add course button on the moderator home page or on list of courses.

yaksh/views.py
```diff
@@ -605,7 +605,7 @@ def course_detail(request, course_id):
     user = request.user
     if not is_moderator(user):
         raise Http404('You are not allowed to view this page')
-    course = get_object_or_404(Course, creator=user, pk=course_id)
+    course = Course.objects.filter(Q(creator=user)|Q(teachers=user), pk=course_id).first()
```
prathamesh920: If user is not a moderator an exception is raised. Similarly an exception should be raised if the user is not allotted to the requested course.
In this case, I can copy-paste the url. Though course will be None, I can access the page.
So I can view other moderator quizzes as well.

yaksh/views.py
```diff
@@ -954,3 +956,75 @@ def design_questionpaper(request):
...
+    course = get_object_or_404(Course, creator=user, pk=course_id)
+    context['course'] = course
+
+    if request.method == 'POST':
+        u_name = request.POST.get('uname')
+        if len(u_name) == 0:
+            return my_render_to_response('yaksh/addteacher.html', context,
+                                        context_instance=ci)
+        else:
+            teachers = User.objects.filter(Q(username__icontains=u_name)|
```
prathamesh920: Everytime I search I will get the users who are already added to the course.
Can you filter the result before displaying

prathamesh920:
Where can a moderator view the teachers in his course?
Also, he should be able to remove the teachers from the course.

adityacp:
added_teachers
search_teacher
show_course

adityacp:
All the changes are resolved

yaksh/views.py
```diff
@@ -608,7 +610,10 @@ def course_detail(request, course_id):
     ci = RequestContext(request)
     if not is_moderator(user):
         raise Http404('You are not allowed to view this page')
-    course = get_object_or_404(Course, creator=user, pk=course_id)
+    course = Course.objects.filter(Q(creator=user)|Q(teachers=user),
```
prathamesh920: If in case course is None then next statement course.creator will throw an error.

yaksh/views.py
```diff
@@ -1025,3 +1036,92 @@ def edit_profile(request):
...
+        context['teachers_rejected'] = teachers_rejected
+        return my_render_to_response('yaksh/addteacher.html', context,
+                                    context_instance=ci)
+    else:
+        return my_render_to_response('yaksh/addteacher.html', context,
+                                    context_instance=ci)
+
+
+@login_required
+def view_courses(request):
```
prathamesh920: I think instead of view_courses it can be allotted_courses.
Also, in template you have viewcourses, which can be allotted_courses.

yaksh/views.py
```diff
@@ -619,7 +624,8 @@ def enroll(request, course_id, user_id=None, was_rejected=False):
     ci = RequestContext(request)
     if not is_moderator(user):
         raise Http404('You are not allowed to view this page')
-    course = get_object_or_404(Course, creator=user, pk=course_id)
+    course = Course.objects.filter(Q(creator=user)|Q(teachers=user),
```
prathamesh920: Here, can we get the code as per the id.
And then call course.is_creator or course.is_teacher?
We can do this for 
if quiz.course.creator != user:
as well.

yaksh/models.py
```diff
@@ -116,6 +124,25 @@ def activate(self):
     def deactivate(self):
         self.active = False
 
+    def add_teachers(self, *teachers):
```
prathamesh920: If feasible then try:
add_teacher only takes teachers with profile, so the views will call profile hasattr which will give teachers with profile to be added. And then call to the add_teacher.

prathamesh920:
You have not address some comments, please revert to those comments.

adityacp:
Can you please specify which comments I missed.

yaksh/views.py
adityacp: resolved

yaksh/models.py
adityacp: resolved

adityacp:
changes resolved

yaksh/views.py
```diff
@@ -69,6 +70,13 @@ def has_profile(user):
     """ check if user has profile """
     return True if hasattr(user, 'profile') else False
 
+def add_teachers_to_group(teachers):
```
prathamesh920: instead you can name it as add_to_group
And instead of teachers as parameter, you can say users.

adityacp: resolved

prathamesh920:
Kindly undo the last blank commit, otherwise looks good.

adityacp:
resolved

PR #95, opened 2016-04-26
Grade User

maheshgudi:
Changed design for grade user.


yaksh/views.py
```diff
@@ -871,36 +871,48 @@ def download_csv(request, questionpaper_id):
...
+                papers = data['papers']
+                for paper in papers:
+                    for question, answers in paper.get_question_answers().iteritems():
+                        marks = float(request.POST.get('q%d_marks' % question.id, 0))
+                        last_ans = answers[-1]
+                        last_ans.marks = marks
+                        last_ans.save()
+                    paper.comments = request.POST.get(
+                        'comments_%d' % paper.question_paper.id, 'No comments')
+                    paper.save()
```
prathamesh920: You will need to update marks obtained, percentage and status for answerpaper  before saving. 
Then only it will reflect on the answerpaper details.

yaksh/views.py
```diff
@@ -871,36 +871,48 @@ def download_csv(request, questionpaper_id):
...
-
-        context = {'data': data}
-        return my_render_to_response('yaksh/user_data.html', context,
-                                     context_instance=ci)
-    else:
-        context = {'data': data}
-        return my_render_to_response('yaksh/grade_user.html', context,
-                                     context_instance=ci)
+
+    course_details =Course.objects.filter(creator = current_user)
```
prathamesh920: space after "=" missing.  Also at line number 892.
Kindly use pep8.

yaksh/views.py
```diff
@@ -730,16 +730,17 @@ def monitor(request, questionpaper_id=None):
...
     """For a given username, this returns a dictionary of important data
     related to the user including all the user's answers submitted.
     """
-    user = User.objects.get(username=username)
+    user = User.objects.get(id=user_id)
     papers = AnswerPaper.objects.filter(user=user)
-    if questionpaper_id is not None:
-        papers = papers.filter(question_paper_id=questionpaper_id).order_by(
+    if questionpaper_id and attempt_number is not None:
+        papers = papers.filter(question_paper_id=questionpaper_id, attempt_number = attempt_number).order_by(
```
prathamesh920: should be 80 cols

prathamesh920: why do we need order_by here as we are passing attempt_number and questionpaper_id?
I think the query will return only one paper.

yaksh/views.py
```diff
@@ -871,36 +871,48 @@ def download_csv(request, questionpaper_id):
...
+    if quiz_id != None:
+        questionpaper_id = QuestionPaper.objects.filter(quiz_id = quiz_id).values("id")
+        user_details = AnswerPaper.objects.get_users_for_questionpaper(questionpaper_id)
+        context = {"users": user_details, "quiz_id": quiz_id}
+        
+        if user_id !=None:
+            attempts = AnswerPaper.objects.get_user_all_attempts(questionpaper_id, user_id)
+            
+            if attempt_number == None:
+                attempt_number = attempts[0].attempt_number
```
prathamesh920: What if attempts is None over here?

yaksh/views.py
```diff
@@ -871,36 +871,48 @@ def download_csv(request, questionpaper_id):
...
+                
+            data = get_user_data(user_id, questionpaper_id, attempt_number)
+            if request.method == "POST":
+                papers = data['papers']
+                for paper in papers:
+                    for question, answers in paper.get_question_answers().iteritems():
+                        marks = float(request.POST.get('q%d_marks' % question.id, 0))
+                        last_ans = answers[-1]
```
prathamesh920: Can we have a custom method update_marks for Answer model?
So we can simply do answer.update_marks(marks)

yaksh/views.py
```diff
@@ -730,16 +730,17 @@ def monitor(request, questionpaper_id=None):
     return my_render_to_response('yaksh/monitor.html', context,
                                  context_instance=ci)
 
-
-def get_user_data(username, questionpaper_id=None):
+def get_user_data(user_id, questionpaper_id=None, attempt_number = None):
```
prathamesh920: Instead of get_user_data in the views, can we have it in the Answerpaper object manager?

yaksh/views.py
```diff
@@ -871,36 +871,48 @@ def download_csv(request, questionpaper_id):
...
+        context = {"users": user_details, "quiz_id": quiz_id}
+        
+        if user_id !=None:
+            attempts = AnswerPaper.objects.get_user_all_attempts(questionpaper_id, user_id)
+            
+            if attempt_number == None:
+                attempt_number = attempts[0].attempt_number
+                
+            data = get_user_data(user_id, questionpaper_id, attempt_number)
+            if request.method == "POST":
```
prathamesh920: you can have ```if request.method == "POST":``` at the end out of the other if statements.

yaksh/views.py

yaksh/views.py

yaksh/views.py

yaksh/views.py

yaksh/views.py

yaksh/views.py

yaksh/views.py

yaksh/views.py
```diff
@@ -873,34 +855,53 @@ def download_csv(request, questionpaper_id):
...
+                                                        .values("id")
+        user_details = AnswerPaper.objects.get_users_for_questionpaper\
+                                            (questionpaper_id)
+        context = {"users": user_details, "quiz_id": quiz_id}
+        if user_id is not None:
+            try:
+                attempts = AnswerPaper.objects.get_user_all_attempts\
+                                                (questionpaper_id, user_id)
+                if attempt_number is None:
+                    attempt_number = attempts[0].attempt_number
```
prathamesh920: should not try be inside if statement?

yaksh/models.py
```diff
@@ -486,6 +496,33 @@ def get_user_answerpapers(self, user):
...
+            .distinct()
+
+    def get_user_all_attempts(self, questionpaper, user):
+        return self.filter(question_paper=questionpaper, user=user)\
+                            .order_by('-attempt_number')
+
+    def get_user_data(self, user, questionpaper_id, attempt_number = None):
+
+        if attempt_number is not None:
+            papers = self.filter( user=user, question_paper_id=questionpaper_id,
```
prathamesh920: please keep consistency (wherever necessary) in the space around assignment operator '=' also after comma

yaksh/tests.py
```diff
@@ -375,6 +380,11 @@ def test_get_previous_answers(self):
         self.assertEqual(answers.count(), 1)
         self.assertTrue(answers[0], self.answer_wrong)
 
+    def test_set_marks (self):
+        self.answer_wrong.set_marks(0.5)
+        self.assertEqual(self.answer_wrong.marks, 0.5)
```
prathamesh920: Also you can test if you allot marks greater then the max marks.

prathamesh920:
Please resolve the comments otherwise looks fine

yaksh/views.py
```diff
@@ -873,34 +855,53 @@ def download_csv(request, questionpaper_id):
...
+                    }   
+    if request.method == "POST":
         papers = data['papers']
         for paper in papers:
             for question, answers in paper.get_question_answers().iteritems():
                 marks = float(request.POST.get('q%d_marks' % question.id, 0))
-                last_ans = answers[-1]
-                last_ans.marks = marks
-                last_ans.save()
+                answers = answers[-1]
```
prathamesh920: If possible to do then try else its fine:
It is always better to have a function in models or model manger
For example:
instead of answer[-1] better would be Answers.objects.get_last_answer(question_id)

yaksh/tests.py
maheshgudi: resolved

yaksh/views.py
maheshgudi: This would require refactoring of get_question_answers() since it returns answers as a list. I can do it in the future updates.

yaksh/models.py
maheshgudi: resolved.

yaksh/views.py

yaksh/tests.py
maheshgudi: resolved

maheshgudi:
sending a clean pr

PR #99, opened 2016-05-07
Fix timezone

adityacp:
user can set timezone and can view time as per timezone

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
 install:
   - pip install git+https://github.com/FOSSEE/online_test.git#egg=yaksh-0.1
   - pip install -q Django==$DJANGO --use-mirrors
+  - pip install -q pytz
 
 before_install:
   - sudo apt-get update -qq
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
 install:
   - pip install git+https://github.com/FOSSEE/online_test.git#egg=yaksh-0.1
   - pip install -q Django==$DJANGO --use-mirrors
-  - pip install -q pytz
+  - pip install -q pytz==2016.4
 
 before_install:
   - sudo apt-get update -qq
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 django==1.9.5,
 mysql-python==1.2.5,
 django-taggit==0.18.1,
+pytz==2016.4
--- a/yaksh/forms.py
+++ b/yaksh/forms.py
@@ -11,6 +11,7 @@ from django.forms.models import inlineformset_factory
 from django.db.models import Q
 from string import letters, punctuation, digits
 import datetime
+import pytz
 
 languages = (
     ("select", "Select Language"),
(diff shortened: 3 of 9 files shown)


.travis.yml
```diff
@@ -11,6 +11,7 @@ env:
 install:
   - pip install git+https://github.com/FOSSEE/online_test.git#egg=yaksh-0.1
   - pip install -q Django==$DJANGO --use-mirrors
+  - pip install -q pytz
```
ankitjavalkar: Please check the version that you have installed while testing the feature and use the same one here,
```pip install -q pytz==<version>```
(instead of <version> use the version number that you installed)

File: requirements.txt [hunk 2]
```diff
@@ -1,3 +1,4 @@
 django==1.9.5,
 mysql-python==1.2.5,
 django-taggit==0.18.1,
+pytz==2016.4
```
ankitjavalkar: Is 2016.4 the version number? Please use a version number here and use the same version number in the travis.yml

yaksh/forms.py
```diff
@@ -196,7 +199,7 @@ class ProfileForm(forms.ModelForm):
     class Meta:
         model = Profile
         fields = ['first_name', 'last_name', 'institute',
-                'department', 'roll_number', 'position']
+                'department', 'roll_number', 'position', 'timezone']
```
ankitjavalkar: Nothing wrong here, just a suggestion, use ```exclude = ['fieldname']``` so that you don't need to add a field everytime you make some change.

yaksh/forms.py
```diff
@@ -57,6 +58,7 @@ class UserRegisterForm(forms.Form):
                 (max_length=64, help_text='Department you work/study at')
     position = forms.CharField\
         (max_length=64, help_text='Student/Faculty/Researcher/Industry/etc.')
+    timezone = forms.ChoiceField(choices=[(tz, tz) for tz in pytz.common_timezones])
```
ankitjavalkar: Please set the default as UTC here, in case someone does not fill the Profile form or the Timezone field

requirements.txt
adityacp: 2016.4 is the version number for pytz

ankitjavalkar: OK

ankitjavalkar:
Looks good, Mergeable.

yaksh/models.py
```diff
@@ -47,6 +48,9 @@
                 ('completed', 'Completed'),
               )
 
+# get current timezone info
+tz = pytz.timezone(timezone.get_current_timezone_name())
```
prabhuramachandran: This makes no sense?  Why is this server-side global being used at all?

yaksh/models.py
```diff
@@ -302,7 +299,9 @@ class Quiz(models.Model):
 
     # The end date and time of the quiz
     end_date_time = models.DateTimeField("End Date and Time of the quiz",
-                                        default=datetime(2199, 1, 1, tzinfo=tz),
+                                        default=datetime(2199, 1, 1,
+                                        tzinfo=pytz.timezone\
```
prabhuramachandran: PEP 8, this would be nicer if this were done as discussed.

PR #101, opened 2016-05-11
Test modes

maheshgudi:
PR for test-modes which includes:
1. Modes for testing quizzes - (God and normal user mode)
2. mode for testing single and multiple selected questions

### DIFF
--- a/yaksh/forms.py
+++ b/yaksh/forms.py
@@ -204,7 +204,6 @@ class CourseForm(forms.ModelForm):
     class Meta:
         model = Course
         fields = ['name', 'active', 'enrollment']
-        exclude = ["is_trial"]
 
 class ProfileForm(forms.ModelForm):
     """ profile form for students and moderators """
--- a/yaksh/forms.py
+++ b/yaksh/forms.py
@@ -163,7 +163,7 @@ class QuizForm(forms.ModelForm):
 
     class Meta:
         model = Quiz
-        fields = '__all__'
+        exclude = ["is_trial"]
 
 
 class QuestionForm(forms.ModelForm):
@@ -204,7 +204,7 @@ class CourseForm(forms.ModelForm):
     class Meta:
         model = Course
         fields = ['name', 'active', 'enrollment']
-
+        exclude = ["is_trial"]
 
 class ProfileForm(forms.ModelForm):
     """ profile form for students and moderators """
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -303,7 +303,7 @@ class QuizManager(models.Manager):
         """Creates a trial quiz for testing questions"""
         trial_quiz = self.create(course=trial_course,
                                  duration=1000,
-                                 description="trial_quiz",
+                                 description="trial_questions",
                                  is_trial=True,
                                  time_between_attempts=0
                                  )
@@ -311,9 +311,11 @@ class QuizManager(models.Manager):
 
     def create_trial_from_quiz(self, original_quiz_id, user, godmode):
         """Creates a trial quiz from existing quiz"""
-        trial_quiz_name = "trial_orig_id_{0}".format(original_quiz_id)
+        trial_quiz_name = "Trial_orig_id_{0}_{1}".format(original_quiz_id,
+                                                        "godmode" if godmode else "usermode"
+                                                        )
         
-        if self.filter(description=trial_quiz_name).exists() and not godmode:
+        if self.filter(description=trial_quiz_name).exists():
             trial_quiz = self.get(description=trial_quiz_name)
             
         else:
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -314,7 +314,7 @@ class QuizManager(models.Manager):
                                  )
         return trial_quiz
 
-    def create_trial_from_quiz(self, original_quiz_id, user, mode):
+    def create_trial_from_quiz(self, original_quiz_id, user, godmode):
         """Creates a trial quiz from existing quiz"""
         trial_quiz = self.get(id=original_quiz_id)
         trial_quiz.course.enroll(False, user)
@@ -323,7 +323,7 @@ class QuizManager(models.Manager):
         trial_quiz.is_trial = True
         trial_quiz.time_between_attempts = 0
         trial_quiz.prerequisite = None
-        if mode == "godmode":
+        if godmode:
             trial_quiz.duration = 1000
             trial_quiz.active = True
             trial_quiz.start_date_time = datetime.now()
@@ -459,8 +459,6 @@ class QuestionPaper(models.Model):
 
     # Total marks for the question paper.
     total_marks = models.FloatField()
-
-#    is_trial = models.BooleanField(default=False)
 
     objects = QuestionPaperManager()
 
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -69,11 +69,6 @@ class CourseManager(models.Manager):
         trial_course.enroll(False, user)
         return trial_course
 
-    def delete_all_trial_courses(self, user):
-        """Deletes all trial course for a user."""
-        trial_course = self.filter(creator=user, is_trial=True)
-        trial_course.delete()
-
 
 ###############################################################################
 class Course(models.Model):
@@ -316,27 +311,26 @@ class QuizManager(models.Manager):
 
     def create_trial_from_quiz(self, original_quiz_id, user, godmode):
         """Creates a trial quiz from existing quiz"""
-        trial_quiz = self.get(id=original_quiz_id)
-        trial_quiz.course.enroll(False, user)
-        trial_quiz.pk = None
-        trial_quiz.description += "_trial"
-        trial_quiz.is_trial = True
-        trial_quiz.time_between_attempts = 0
-        trial_quiz.prerequisite = None
-        if godmode:
-            trial_quiz.duration = 1000
-            trial_quiz.active = True
-            trial_quiz.start_date_time = datetime.now()
-            trial_quiz.end_date_time = datetime(2199, 1, 1, 0, 0, 0, 0)
-        trial_quiz.save()
+        trial_quiz_name = "trial_orig_id_{0}".format(original_quiz_id)
+        
+        if self.filter(description=trial_quiz_name).exists() and not godmode:
+            trial_quiz = self.get(description=trial_quiz_name)
+            
+        else:
+            trial_quiz = self.get(id=original_quiz_id)
+            trial_quiz.course.enroll(False, user)
+            trial_quiz.pk = None
+            trial_quiz.description = trial_quiz_name
+            trial_quiz.is_trial = True
+            trial_quiz.time_between_attempts = 0
+            trial_quiz.prerequisite = None
+            if godmode:
+                trial_quiz.duration = 1000
+                trial_quiz.active = True
+                trial_quiz.start_date_time = timezone.now()
+                trial_quiz.end_date_time = datetime(2199, 1, 1, 0, 0, 0, 0)
+            trial_quiz.save()
         return trial_quiz
-
-    def delete_all_trial_quizzes(self, user):
-        trial_quiz = self.filter(Q(course__creator=user) |
-                                 Q(course__teachers=user), is_trial=True
-                                 )
-        trial_quiz.delete()
-
 
 ###############################################################################
 class Quiz(models.Model):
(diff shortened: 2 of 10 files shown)


prabhuramachandran:
Please add tests for any model changes you have made.

yaksh/models.py
```diff
@@ -58,6 +59,23 @@ def get_model_class(model):
...
+        trial_course = self.create(name="trial_course",
+                                    enrollment="open",
+                                    creator=user,
+                                    is_trial=True
+                                    )
+        was_rejected=False
+        trial_course.enroll(was_rejected,user)
+        return trial_course
+
+    def delete_trial_course(self, user):
```
prabhuramachandran: Shouldn't this  really be `delete_all_trial_courses`?

yaksh/models.py
```diff
@@ -280,7 +299,40 @@ def __unicode__(self):
...
+        trial_quiz = self.create(course=trial_course,
+                                 duration=1000,
+                                 description="trial_quiz",
+                                 is_trial=True,
+                                 time_between_attempts=0
+                                )
+        return trial_quiz
+
+    def copy_original_quiz(self, original_quiz_id, user, mode):
+        was_rejected = False
```
prabhuramachandran: Why is this variable needed?

yaksh/models.py
```diff
@@ -280,7 +299,40 @@ def __unicode__(self):
...
+        trial_quiz = self.get(id=original_quiz_id)
+        trial_quiz.course.enroll(was_rejected,user)
```
prabhuramachandran: PEP8 please.

yaksh/models.py
```diff
@@ -280,7 +299,40 @@ def __unicode__(self):
...
+    def create_trial_quiz(self,trial_course,user):
+        trial_quiz = self.create(course=trial_course,
+                                 duration=1000,
+                                 description="trial_quiz",
+                                 is_trial=True,
+                                 time_between_attempts=0
+                                )
+        return trial_quiz
+
+    def copy_original_quiz(self, original_quiz_id, user, mode):
```
prabhuramachandran: `create_trail_from_quiz` may be a better name?

yaksh/models.py
```diff
@@ -280,7 +299,40 @@ def __unicode__(self):
...
+            trial_quiz.duration = 1000
+            trial_quiz.active = True
+            trial_quiz.start_date_time = datetime.now()
+            trial_quiz.end_date_time = datetime(2199, 1, 1, 0, 0, 0, 0)
+            trial_quiz.time_between_attempts = 0
+        trial_quiz.save()
+        return trial_quiz    
+
+    def delete_trial_quiz(self, user):
+        trial_quiz = self.filter(Q(course__creator=user)|Q(course__teachers=user),
```
prabhuramachandran: will this work when a course has multiple teachers?  Add a test case for this.

yaksh/models.py
```diff
@@ -335,14 +389,53 @@ def is_expired(self):
...
+        trial_questionpaper.save()
+        return trial_questionpaper, trial_questions
+
+    def _create_trial_questionpaper(self, trial_quiz):
+        trial_questionpaper=self.create(quiz=trial_quiz,
+                                        total_marks=10,
+                                        )
+        return trial_questionpaper
+
+    def add_details_trial_questionpaper(self, trial_quiz,
```
prabhuramachandran: Can't this be `create_trial_paper` with suitable arguments with a proper docstring?

yaksh/models.py
maheshgudi: Yes, it does work for multiple teachers.

yaksh/models.py
maheshgudi: resolved

yaksh/models.py
maheshgudi: resolved

yaksh/models.py
maheshgudi: resolved

yaksh/models.py
maheshgudi: resolved. changed it to create_trial_from_quiz

yaksh/models.py
maheshgudi: resolved. split the "add_details" method in two separate and added docstrings.

maheshgudi:
Added test cases for all test_mode related methods

yaksh/forms.py
```diff
@@ -200,7 +200,7 @@ class CourseForm(forms.ModelForm):
     class Meta:
         model = Course
         fields = ['name', 'active', 'enrollment']
-
+        exclude = ["is_trial"]
```
ankitjavalkar: Suggestion - Not exactly needed since fields already will only show the above mentioned fields

ankitjavalkar:
looks good

prathamesh920:
Please use PEP8

yaksh/models.py
```diff
@@ -366,6 +460,10 @@ class QuestionPaper(models.Model):
     # Total marks for the question paper.
     total_marks = models.FloatField()
 
+#    is_trial = models.BooleanField(default=False)
```
prabhuramachandran: Please remove commented code.

yaksh/static/yaksh/js/add_quiz.js
```diff
@@ -9,3 +9,22 @@ function test()
...
+String.prototype.beginsWith = function (string) {
+    return(this.indexOf(string) === 0);
+};
+
+function usermode(location)
+{
+  var select = document.getElementById("id_prerequisite");
+  var select_text = select.options[select.selectedIndex].text;
+  if (select_text.beginsWith("----")){
+  		window.alert("No prerequisite for this course.\n \
```
prabhuramachandran: Not sure if window.alert is the best way to do this, perhaps this can be changed later to use notify.js or some other library.

yaksh/models.py
```diff
@@ -280,10 +298,46 @@ def set_marks(self, marks):
...
+        """Creates a trial quiz for testing questions"""
+        trial_quiz = self.create(course=trial_course,
+                                 duration=1000,
+                                 description="trial_quiz",
+                                 is_trial=True,
+                                 time_between_attempts=0
+                                 )
+        return trial_quiz
+
+    def create_trial_from_quiz(self, original_quiz_id, user, mode):
```
prabhuramachandran: Can you make `mode`, `godmode=False` as it appears you only have two modes user mode and god mode.   This way instead of a string argument it is a boolean which defaults to False.

yaksh/tests.py
```diff
@@ -182,6 +187,71 @@ def test_get_active_quizzes(self):
...
+        self.assertEqual(trial_quiz.duration, 1000)
+        self.assertEqual(trial_quiz.description, "trial_quiz")
+        self.assertTrue(trial_quiz.is_trial)
+        self.assertEqual(trial_quiz.time_between_attempts, 0)
+
+    def test_create_trial_from_quiz_godmode(self):
+        """Test to check if a copy of original quiz is created in godmode"""
+        trial_quiz = Quiz.objects.create_trial_from_quiz(self.quiz1.id,
+                                                         self.creator,
+                                                         "godmode"
```
prabhuramachandran: Here this would simply become `godmode=True`

yaksh/views.py
```diff
@@ -580,7 +595,7 @@ def add_course(request):
 def enroll_request(request, course_id):
     user = request.user
     ci = RequestContext(request)
-    course = get_object_or_404(Course, pk=course_id)
+    course = get_object_or_404(Course, pk=course_id, is_trial=False)
```
prabhuramachandran: This is repeated all over, can there be a convenient function to do this nicely?  The need to set `is_trial=False` everywhere seems a bit disturbing.

prabhuramachandran: For example, could one make the `is_trial=False` the default and set it to `True` only when needed?

yaksh/models.py
maheshgudi: resolved

yaksh/views.py
maheshgudi: There are two ways to go about this -
1. We can override django's queryset api and force it to always return only a ```is_trial=False``` queryset object. But the problem is its not a good thing to tinker around django's api. 
2. We can create a manager method but it seems quite unnecessary since most of the ```is_trial=False``` filter calls have been made only once in views.

So I have kept some ```is_trial``` while a few unecessary ones I have deleted.

yaksh/views.py
```diff
@@ -281,7 +284,11 @@ def prof_manage(request):
 rights/permissions and log in."""
     user = request.user
     if user.is_authenticated() and is_moderator(user):
-        question_papers = QuestionPaper.objects.filter(quiz__course__creator=user)
+        question_papers = QuestionPaper.objects.filter(quiz__course__creator=user,
+                                                       quiz__is_trial=False
+                                                       )
+        trial_course = Course.objects.delete_all_trial_courses(user)
```
prathamesh920: Is it better to give manual delete?
Because then the moderator can see the attempt as well.
Also, I guess in the user mode, moderator can attempt infinite time even though number of attempt is set to one, because previous attempt is deleted!

yaksh/views.py
```diff
@@ -835,6 +851,13 @@ def show_all_questions(request):
                 msg = "Please select atleast one question"
                 context['msg'] = msg
 
+        if request.POST.get('test') == 'test':
+            question_ids = request.POST.getlist("question")
```
prathamesh920: If "question" is not selected and I click "test" then will throw error.

yaksh/views.py
```diff
@@ -1176,3 +1200,29 @@ def remove_teachers(request, course_id):
...
+    return trial_questionpaper
+
+
+@login_required
+def test_quiz(request, mode, quiz_id):
+    """creates a trial quiz for the moderators"""
+    godmode = True if mode == "godmode" else False
+    current_user = request.user
+    trial_questionpaper = test_mode(current_user, godmode, None, quiz_id)
+    return my_redirect("/exam/start/{0}".format(trial_questionpaper.id))
```
prathamesh920: After testing questions or taking trial_quiz, moderator is redirected to his dashboard.
It will be good if you redirect to the same page where we click "test",  "user mode" or "god mode" buttons.

yaksh/templates/yaksh/add_quiz.html
```diff
@@ -11,15 +11,41 @@
...
+    <a href="" onclick="$('#help').show(); return false;">Help </a>
+    </center>
+    <br/>
+   <div style="display: none;" id="help">
+   <ol>
+   <li><b>User Mode:</b> Attempt quiz the way normal users will attempt i.e. -
+   <ul>
+   <li><i>Quiz will have the same duration as that of the original quiz.</li>
+   <li>Quiz won't start if the course is inactive or the quiz time has expired.</li>
+   <li>You will be notified about quiz prerequisites.(You can still attempt the quiz though)</i></li>
```
prathamesh920: In User mode we can attempt quiz without passing prequisite.
This is already there for God mode, so User mode should be completely same as it is for a student.

yaksh/views.py
```diff
@@ -1176,3 +1200,29 @@ def remove_teachers(request, course_id):
...
+    else:
+        trial_quiz = Quiz.objects.create_trial_from_quiz(quiz_id, user, godmode)
+        trial_questionpaper = QuestionPaper.objects\
+                                           .create_trial_paper_to_test_quiz\
+                                            (trial_quiz, quiz_id)
+    return trial_questionpaper
+
+
+@login_required
+def test_quiz(request, mode, quiz_id):
```
prathamesh920: If quiz is expired or inactive then it creates trial question paper and then redirects to start, then
it gives message that quiz is expired or inactive, then redirects.
Better if it directly gives message that the quiz is expired or inactive.

yaksh/models.py
```diff
@@ -70,7 +87,8 @@ class Course(models.Model):
     rejected = models.ManyToManyField(User, related_name='rejected')
     created_on = models.DateTimeField(auto_now_add=True)
     teachers = models.ManyToManyField(User, related_name='teachers')
-
+    is_trial = models.BooleanField(default=False)
```
prathamesh920: Nice, default value given. 
@all 
Always give default value to the new attribute added, this makes database migration easy.

yaksh/views.py

yaksh/templates/yaksh/add_quiz.html
maheshgudi: resolved

yaksh/views.py

prathamesh920:
This PR is mergeable except we need to fix [Issue #109  "User Mode Testing: Prerequisite Test Relation issue"](https://github.com/FOSSEE/online_test/issues/109)

PR #105, opened 2016-06-02
Stdin stdout evaluators

adityacp:
Created Stdio based evaluators for C, Cpp, Java, Python and bash.
Evaluators takes the expected input and expected output provided by moderator
Compares expected output with user output 
Added test cases for evaluators


yaksh/bash_stdio_evaluator.py
```diff
@@ -0,0 +1,44 @@
...
+    def teardown(self):
+        super(BashStdioEvaluator, self).teardown()
+        os.remove(self.submit_code_path)
+
+    def compile_code(self, user_answer, expected_input, expected_output):
+        if not isfile(self.submit_code_path):
+            msg = "No file at %s or Incorrect path" % self.submit_code_path
+            return False, msg
+        user_code_directory = os.getcwd() + '/'
+        user_answer = user_answer.replace("\r", "")
```
prathamesh920: Is it necessary to replace?

prathamesh920:
cpp_stdio_evaluator.py and cpp_code_evaluator.py both files are almost similar (redundancy).
Similarly, java_stdio_evaluator and java_code_evaluator are similar.
Is it possible to  remove the redundancy?

adityacp:
Redundancy of two files can be removed but will take time in testing...for now this works properly

yaksh/bash_stdio_evaluator.py
adityacp: It is necessary to replace '\r' because the code while executing generates error "command not found".

yaksh/bash_stdio_evaluator.py
```diff
@@ -0,0 +1,44 @@
...
+        self.write_to_submit_code_file(self.submit_code_path, user_answer)
+
+    def check_code(self, user_answer, expected_input, expected_output):
+        success = False
+        expected_input = str(expected_input).replace('\r', '')
+        proc = subprocess.Popen(["bash ./Test.sh"],
+                                shell=True,
```
prathamesh920: If shell is True so better to have the command "bash ./Test.sh" as a string instead of sequence.

yaksh/python_stdio_evaluator.py
```diff
@@ -0,0 +1,63 @@
...
+            exec_scope = {}
+            exec submitted in exec_scope
+        self.output_value = output_buffer.getvalue().rstrip("\n")
+        return self.output_value
+
+    def check_code(self, user_answer, expected_input, expected_output):
+        success = False
+
+        tb = None
+        if expected_output in user_answer:
```
prathamesh920: checking this expected_output in user_answer may give unexpected results.
Better to remove it because even though my answer code is correct but will still get incorrect, only because I have expected_output some where in my code.
To check whether user is simply printing, we have multiple testcases to take care of this situation for now.

yaksh/cpp_stdio_evaluator.py
```diff
@@ -0,0 +1,105 @@
...
+            err = "Compilation Error:"
+            try:
+                error_lines = stdnt_stderr.splitlines()
+                for e in error_lines:
+                    if ':' in e:
+                        err = err + "\n" + e.split(":", 1)[1]
+                    else:
+                        err = err + "\n" + e
+            except:
+                err = err + "\n" + stdnt_stderr
```
prathamesh920: We can have a method for error handling, which will produce error in a proper format. This can be done later. For now I am raising an issue.

yaksh/evaluator_tests/test_bash_evaluation.py
```diff
@@ -48,5 +51,81 @@ def test_infinite_loop(self):
...
+        self.assertTrue(result.get('success'))
+
+    def test_incorrect_answer(self):
+        user_answer = dedent(""" #!/bin/bash
+                             read A
+                             read B
+                             echo -n `expr $A + $B`
+                             """
+                             )
+        test_case_data = [{'expected_output': '12', 'expected_input': '5\n6'}]
```
prathamesh920: This testcase is fine but would be better if you write a wrong code, like subtraction instead of addition.
The code is proper but expected output is something else, a bit not clear.
Expected output can be 11 and not 12.

yaksh/evaluator_tests/test_bash_evaluation.py
```diff
@@ -48,5 +51,81 @@ def test_infinite_loop(self):
         self.assertEquals(result.get("error"), self.timeout_msg)
 
 
+class BashStdioEvaluationTestCases(unittest.TestCase):
+    def setUp(self):
+        self.timeout_msg = ("Code took more than {0} seconds to run. "
```
prathamesh920: It will be better if you have a testcase for infinite loop.

yaksh/evaluator_tests/test_c_cpp_evaluation.py
```diff
@@ -52,5 +55,264 @@ def test_infinite_loop(self):
...
+        cout<<a+b;
+        }""")
+        get_class = CppStdioEvaluator()
+        kwargs = {'user_answer': user_answer,
+                  'test_case_data': self.test_case_data
+                  }
+        result = get_class.evaluate(**kwargs)
+        self.assertEquals(result.get('error'), "Correct Answer")
+        self.assertTrue(result.get('success'))
+
```
prathamesh920: To test c and cpp we have two different classes, performing same function, only the answer code changes, can we have a single class to test both?

yaksh/evaluator_tests/test_java_evaluation.py
```diff
@@ -61,5 +63,128 @@ def test_infinite_loop(self):
...
+         System.out.print(a+b);
+        }}""")
+        get_class = JavaStdioEvaluator()
+        kwargs = {'user_answer': user_answer,
+                  'test_case_data': self.test_case_data
+                  }
+        result = get_class.evaluate(**kwargs)
+        self.assertEquals(result.get('error'), "Correct Answer")
+        self.assertTrue(result.get('success'))
+
```
prathamesh920: Can we have a testcase to check string input/output?

yaksh/evaluator_tests/test_python_evaluation.py
```diff
@@ -164,71 +164,166 @@ def add(a, b):
...
+        kwargs = {'user_answer': user_answer,
+                  'test_case_data': self.test_case_data
+                  }
+        result = get_class.evaluate(**kwargs)
+        self.assertTrue(result.get('success'))
+        self.assertIn("Correct Answer", result.get('error'))
+
+    def test_add_two_integers_incorrect(self):
+        self.test_case_data = [{"expected_input": "1\n2",
+                                "expected_output": "4"
```
prathamesh920: Again expected output can be 3 and code may print the difference.

yaksh/python_stdio_evaluator.py
prabhuramachandran: Yes this must be done carefully and also properly tested.

yaksh/stdio_evaluator.py
```diff
@@ -0,0 +1,22 @@
+class Evaluator(object):
+
+    def evaluate(self, user_answer, proc, expected_input, expected_output):
```
ankitjavalkar: add a default setup() and teardown() here which is overridden by the child classes

ankitjavalkar:
Looks good, mergeable

ankitjavalkar:
Please sync with new changes (git pull)

PR #106, opened 2016-06-06
Fix timezone

adityacp:
changed naive datetime objects to timezone aware datetime objects to avoid runtime warnings.


ankitjavalkar:
This is mergeable

prathamesh920:
Looks good to me.

PR #110, opened 2016-07-07
Minor change in requirements.txt | removed comma (,)

manish-kr:
While running the command `pip install -r requirements.txt` the following error is observed:
"InvalidRequirement: Invalid requirement, parse error at "u',' "

Fix:
Removed ',' after each module name in requirements.txt to fix the error.


prabhuramachandran:
Thanks!

PR #112, opened 2016-07-08
oauth implemented

prathamesh920:
User can login via google and facebook.
If user logs in from google and then later the same user
logs in from facebook, then he will be considered as the same user if
the email registered on facebook is of gmail.
User profile will be created when user logs in via facebook or google.
In yaksh pipeline, user.py is used to create profile. Pipeline settings
is defined that defines the functions to be executed during oauth.
Used bootstrap social and Font Awesome for UI design


prabhuramachandran:
Please rebase.

PR #113, opened 2016-07-08
File based questions

adityacp:
Moderator can set questions based on files for any language


yaksh/bash_code_evaluator.py
```diff
@@ -22,8 +23,11 @@ def teardown(self):
         # Delete the created file.
         super(BashCodeEvaluator, self).teardown()
         os.remove(self.submit_code_path)
+        if self.files_list:
```
prathamesh920: Can be only self.files:

yaksh/bash_code_evaluator.py
```diff
@@ -22,8 +23,11 @@ def teardown(self):
...
+            file_delete = CopyDeleteFiles()
+            file_delete.delete_files(self.files_list)
```
prathamesh920: Instead of delete_files, it can simply be delete()

yaksh/bash_code_evaluator.py
```diff
@@ -53,6 +57,10 @@ def check_code(self, user_answer, test_case):
         clean_ref_code_path, clean_test_case_path = \
             self._set_test_code_file_path(get_ref_path, get_test_case_path)
 
+        self.files_list = []
+        if file_paths:
+            file_copy = CopyDeleteFiles()
+            self.files_list = file_copy.copy_files(file_paths)
```
prathamesh920: copy() instead of copy_files

yaksh/models.py
```diff
@@ -268,6 +276,35 @@ def __unicode__(self):
...
+    question = models.ForeignKey(Question, related_name="question")
+    extract = models.BooleanField(default=False)
+
+    def delete_all_files(self, files):
+        for file in files:
+            if os.path.exists(file.files.path):
+                shutil.rmtree(os.path.dirname(file.files.path))
+            file.delete()
+
+    def delete_selected_files(self, files):
```
prathamesh920: Instead of two delete methods, can't we have a single delete method?

yaksh/models.py
```diff
@@ -268,6 +276,35 @@ def __unicode__(self):
...
+        for file in files:
+            if os.path.exists(file.files.path):
+                os.remove(file.files.path)
+                if os.listdir(os.path.dirname(file.files.path)) == []:
+                    os.rmdir(os.path.dirname(file.files.path))
+            file.delete()
+
+    def extract_files(self, files):
```
prathamesh920: What is this method doing?

yaksh/copy_delete_files.py
```diff
@@ -0,0 +1,29 @@
+import shutil
+import os
+import zipfile
+
+
+class CopyDeleteFiles(object):
+
+    def copy_files(self, file_paths):
```
prabhuramachandran: Why are these methods of a class?  No need for an object when you don't need one.  Just make them functions and also change the module name to utils or something like that.  Why isn't there a docstring, especially when file_paths is expected to be a specific form of data?

yaksh/file_utils.py
```diff
@@ -0,0 +1,33 @@
+import shutil
```
prathamesh920: utils.py will be fine instead of file_utils, just a suggestion.

yaksh/templates/yaksh/question.html
```diff
@@ -142,7 +142,14 @@ <h3 class="brand"><strong>Online Test</h3></strong>
   <h4><u> {{ question.summary }} </u><font class=pull-right>(Marks : {{ question.points }}) </font></h4><br>
   <font size=3 face=arial> {{ question.description|safe }} </font>
   <br><font size=3 face=arial> Language: {{ question.language }} </font><br>
-
+  {% if files %}
+  <h4>Files to download for this question</h4>
+  {% for file in files %}
```
prathamesh920: can we give a download button option which will download all the files in one go?

yaksh/templates/yaksh/add_question.html
```diff
@@ -15,19 +15,27 @@
...
   <tr><td>Description: <td>{{ form.description}}  {{form.description.errors}}
   <tr><td>Tags: <td>{{ form.tags }}
   <tr><td>Snippet: <td>{{ form.snippet }}
   <tr><td> Test Case Type: <td> {{ form.test_case_type }}{{ form.test_case_type.errors }}
-
+  <tr><td> File: <td> {{ upload_form.file_field }}{{ upload_form.file_field.errors }}
+  {% if uploaded_files %}<br><b>Uploaded files:</b><br>Check the box to delete or extract files<br>
+  {% for file in uploaded_files %}
+  <input type="checkbox" name="clear" value="{{file.id}}">&nbsp;delete</input>&nbsp;
+  <input type="checkbox" name="extract" value="{{file.id}}" >{% if file.extract %} dont extract{% else %}
```
prathamesh920: If a upload a file which is not a zip and click on extract then there should be a validation to check whether a file is zip or not.  Else you can handle this when you copy the files in the user directory.

prathamesh920: Is there a way to avoid upload of files with same name?

yaksh/models.py
```diff
@@ -274,6 +283,29 @@ def __unicode__(self):
 
 
 ###############################################################################
+class QuestionsFileUpload(models.Model):
+    files = models.FileField(upload_to=get_upload_dir, blank=True)
```
prathamesh920: can be file?

yaksh/models.py
```diff
@@ -274,6 +283,29 @@ def __unicode__(self):
 
 
 ###############################################################################
+class QuestionsFileUpload(models.Model):
```
prathamesh920: Can we have name as FileUpload?

prathamesh920: Or may be only File, with attributes: path, question and extract?

yaksh/models.py
```diff
@@ -274,6 +283,29 @@ def __unicode__(self):
...
+    files = models.FileField(upload_to=get_upload_dir, blank=True)
+    question = models.ForeignKey(Question, related_name="question")
+    extract = models.BooleanField(default=False)
+
+    def delete_files(self, files):
```
prathamesh920: This can be simply delete instead of delete_files.

yaksh/views.py
```diff
@@ -176,9 +185,24 @@ def edit_question(request, question_id=None):
         raise Http404('No Question Found')
 
     question_instance = Question.objects.get(id=question_id)
-
+    if request.method == "POST" and 'delete_files' in request.POST:
+        remove_files_id = request.POST.getlist('clear')
+        if remove_files_id:
+            files = QuestionsFileUpload.objects.filter(id__in=remove_files_id)
+            file = QuestionsFileUpload()
+            file.delete_files(files)
```
prathamesh920: In models you have created delete_files, so that you can use file.delete() where file is the object instance but here you are creating new file instance and then calling file.delete() with argument again of same instance type.
I think your delete method should be in File Manager in models. So that you can simply do File.objects.delete(files)
Note: Also django queryset provides delete method.

yaksh/views.py
```diff
@@ -176,9 +185,24 @@ def edit_question(request, question_id=None):
...
+        form = FileForm(request.POST, request.FILES)
+        files = request.FILES.getlist('file_field')
+        extract_files_id = request.POST.getlist('extract')
+        if files:
+            for file in files:
+                QuestionsFileUpload.objects.get_or_create(question=question_instance, files=file)
+        if extract_files_id:
+            files = QuestionsFileUpload.objects.filter(id__in=extract_files_id)
+            file = QuestionsFileUpload()
+            file.set_extract_status(files)
```
prathamesh920: Same issue.
If we want a custom model method then it should accessible in following manner:
Example:
course = Course.objects.get(pk=1)
course.deactivate()
deactivate is model method.

yaksh/file_utils.py
```diff
@@ -0,0 +1,33 @@
...
+                files.append(zip_files)
+            unzip.extractall()
+            unzip.close()
+    return files
+
+
+def delete_files(files):
+    """ Delete Files from current directory """
+
+    for content in files:
```
prathamesh920: for file in files:
Is this fine?

prathamesh920:
Can we have a testcase for file does not exist situation?

yaksh/templates/yaksh/add_question.html
adityacp: while extracting files I have added a validation to check if file to be extracted is a zipfile.

adityacp: currently files with same name will be given any random name to distinguish two files.
later we can add a validation to avoid file with same names

yaksh/file_utils.py
adityacp: file_utils is standard name which  resembles file related features. So I feel this is appropriate since here we are only file doing file related functionalities.

yaksh/templates/yaksh/question.html
adityacp: currently student can download files by clicking on link.
we can add a button to download a zip file
I will raise an issue for this.

adityacp:
Resolved some comments added issues for some comments which needs to be done later
Issue File Based Questions #120

PR #115, opened 2016-07-11
minor changes in timezone and change password templates

maheshgudi:
timezone related -
added default value for timezone in profile class 

test mode related - 
changed time between attempts for god mode to be zero

previous templates for forgot password and change password were from django admin. Changed it to suit the yaksh theme


prathamesh920:
Looks good to me.

PR #122, opened 2016-08-08
made UI changes to courses page and teacher can add another teacher

maheshgudi:
1. Moderator can now see courses created and allotted on the same page.
2. Teacher can add another teacher for a course.


yaksh/views.py
```diff
@@ -1196,36 +1198,27 @@ def add_teacher(request, course_id):
...
-    if not is_moderator(user):
-        raise Http404('You are not allowed to view this page!')
-
-    courses = Course.objects.filter(teachers=user)
-    return my_render_to_response('yaksh/courses.html', {'courses': courses},
-                                        context_instance=ci)
-
 
 @login_required
 def remove_teachers(request, course_id):
```
prathamesh920: remove teacher will throw error for:
course = get_object_or_404(Course, creator=user, pk=course_id)
because user will not be the creator if user is a teacher.
So better will be remove "Remove Selected" teachers button for he allotted courses.
Kindly check all features when changes are implemented.

yaksh/views.py
```diff
@@ -1196,36 +1198,27 @@ def add_teacher(request, course_id):
...
     context['course'] = course
 
     if request.method == 'POST':
         teacher_ids = request.POST.getlist('check')
         teachers = User.objects.filter(id__in=teacher_ids)
-        add_to_group(teachers)
-        course.add_teachers(*teachers)
-        context['status'] = True
-        context['teachers_added'] = teachers
+        if not course.creator in teachers:
```
prathamesh920: Alternative:
In search teacher query we can the course creator.
like: query.exclude(Q(id=course.creator.id)
So you don't need to check the condition neither you need to use any alert in the template.

maheshgudi: Resolved.

yaksh/views.py
maheshgudi: Resolved

yaksh/views.py
```diff
@@ -1164,25 +1166,21 @@ def search_teacher(request, course_id):
         raise Http404('You are not allowed to view this page!')
 
     context = {}
-    course = get_object_or_404(Course, creator=user, pk=course_id)
+    course = get_object_or_404(Course, Q(creator=user)|Q(teachers=user), pk=course_id)
```
prathamesh920: This may return more than one object and raise 404

yaksh/views.py
prathamesh920: Is it a good idea to give "Remove Teacher" option to another teacher?

maheshgudi: From a previous discussion with the team and Prabhu Sir, it was decided that a teacher will have all the rights of the course creator. Hence a teacher gets the right to remove and add other teachers.

prathamesh920: Alright, then looks good to merge.

PR #123, opened 2016-08-12
Removed for loop from the question template.

prathamesh920:
Once the code question is correct, user must see the message Correct
Output for two seconds and then url redirection.
But since this was written in a for loop, so
multiple requests almost simultaneously were sent by a single user.
This caused the server to go down. For now removed for loop, which resolves the issue.


prabhuramachandran:
Thanks for catching this! Merging.

PR #125, opened 2016-08-19
Minor Changes: admin.py and template changes

ankitjavalkar:
- Modify admin.py to include Course model
- Modify user.html and complete.html to include id for html elements


ankitjavalkar:
Minor Changes: Added html IDs to certain tags to allow selenium script to detect the tags. 
Since we are deploying the Newest changes on yaksh.fossee.in instance, we will need the above changes to run the selenium script on yaksh.fossee.in.

Hence merging ASAP

PR #127, opened 2016-08-24
Create course fixtures

adityacp:
Created a Fixture for creating Demo Course, Demo Quiz
Created a Management command to create Demo Course


yaksh/management/commands/create_demo_course.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class Command(BaseCommand):
+    help = "Create a Demo Course, Demo Quiz"
+
+    def handle(self, *args, **options):
+        """ Handle command to create demo course """
+        status = create_demo_course()
+        if status:
+            self.stdout.write("Successfully Created")
```
ankitjavalkar: Add an else clause that just prints "Test creation Failed"

yaksh/models.py
```diff
@@ -231,27 +235,32 @@ def consolidate_answer_data(self, user_answer):
...
+                      'files': files}
             questions_dict.append(q_dict)
-
-        return json.dumps(questions_dict, indent=2)
+        question._add_json_to_zip(zip_file, questions_dict)
+        return zip_file_name
 
     def load_from_json(self, questions_list, user):
         questions = json.loads(questions_list)
+        ques = Question()
```
ankitjavalkar: I don't find this being used anywhere in the function?

yaksh/models.py
```diff
@@ -278,6 +287,29 @@ def get_test_case(self, **kwargs):
 
         return test_case
 
+    def _add_and_get_files(self, zip_file):
+        files = FileUpload.objects.filter(question=self)
+        for file in files:
```
ankitjavalkar: A little confusing since you later go on to use file.file.path etc. I would recommend 'for f in files' or 'for fi in files'

yaksh/models.py
```diff
@@ -231,27 +235,32 @@ def consolidate_answer_data(self, user_answer):
...
 
     def load_from_json(self, questions_list, user):
         questions = json.loads(questions_list)
+        ques = Question()
         for question in questions:
             question['user'] = user
+            files = question.pop('files')
             test_cases = question.pop('testcase')
             que, result = Question.objects.get_or_create(**question)
+            que._add_files_to_db(files)
```
ankitjavalkar: file_names

yaksh/models.py
```diff
@@ -278,6 +287,29 @@ def get_test_case(self, **kwargs):
...
+    def _add_and_get_files(self, zip_file):
+        files = FileUpload.objects.filter(question=self)
+        for file in files:
+            zip_file.write(file.file.path, (os.path.basename(file.file.path)))
+        files_list = [os.path.basename(file.file.name) for file in files]
+        return files_list
+
+    def _add_files_to_db(self, files):
+        if files:
+            for file_name in files:
```
ankitjavalkar: for file_name in file_names:

yaksh/models.py
ankitjavalkar: Please add;
if file_names:
    que._add_files_to_db(file_names)

yaksh/models.py
```diff
@@ -278,6 +287,29 @@ def get_test_case(self, **kwargs):
...
 
+    def _add_and_get_files(self, zip_file):
+        files = FileUpload.objects.filter(question=self)
+        for file in files:
+            zip_file.write(file.file.path, (os.path.basename(file.file.path)))
+        files_list = [os.path.basename(file.file.name) for file in files]
+        return files_list
+
+    def _add_files_to_db(self, files):
+        if files:
```
ankitjavalkar: Please move this if towards the function call

yaksh/views.py
```diff
@@ -663,9 +681,11 @@ def courses(request):
     ci = RequestContext(request)
     if not is_moderator(user):
         raise Http404('You are not allowed to view this page')
-    courses = Course.objects.filter(creator=user, is_trial=False)
+    demo_user = User.objects.get(username="demo_user")
```
ankitjavalkar: You will need to consider the possibility that the management command has not been executed. Please add a try/catch or an if/else to handle the case where the demo_user does not exist

yaksh/views.py
```diff
@@ -73,13 +74,30 @@ def is_moderator(user):
...
+        zip_file.extractall()
+
+
+def read_json(json_file, user):
+    question = Question()
+    if os.path.exists(json_file):
+        with open(json_file, 'r') as q_file:
+            questions_list = q_file.read()
+            question.load_questions(questions_list, user)
+            os.remove(json_file)
```
ankitjavalkar: I guess os.remove should be outside of the 'with' block? In case a file read operation fails, then the json_file will not be deleted.

yaksh/views.py
```diff
@@ -73,13 +74,30 @@ def is_moderator(user):
...
+
 def add_to_group(users):
     """ add users to moderator group """
     group = Group.objects.get(name="moderator")
     for user in users:
         if not is_moderator(user):
             user.groups.add(group)
 
+
+def extract_files(questions_file):
```
ankitjavalkar: Is the zip file created in the current directory? If the files are supposed to be discarded at the end then can we create the files in the /tmp/ to avoid permission issue?

yaksh/views.py
```diff
@@ -869,22 +889,24 @@ def show_all_questions(request):
...
             if form.is_valid():
                 questions_file = request.FILES['file']
-                if questions_file.name.split('.')[-1] == "json":
-                    questions_list = questions_file.read()
-                    question = Question()
-                    question.load_from_json(questions_list, user)
+                file_name = questions_file.name.split('.')
+                if file_name[-1] == "zip":
+                    extract_files(questions_file)
+                    read_json("questions_dump.json", user)
```
ankitjavalkar: Is the file in current directory? can we create the files in the /tmp/ to avoid permission issue?

yaksh/test_models.py
```diff
@@ -143,24 +149,33 @@ def test_dump_questions_into_json(self):
         """ Test dump questions into json """
         question = Question()
         question_id = [self.question2.id]
-        questions = json.loads(question.dump_into_json(question_id, self.user2))
+        questions_zip = question.dump_questions(question_id, self.user2)
+        zip_file = zipfile.ZipFile(questions_zip, "r")
+        zip_file.extractall("/tmp/")
```
prabhuramachandran: Never do this, instead use `tempfile` to create a temporary directory and then clear out that directory in `tearDown`.

ankitjavalkar: I had asked Aditya to make this change, though my intention was the same - the file should be extracted/created in a temp directory and then cleared out. I should have made this more clear.

yaksh/views.py
prathamesh920: Can we have this method in file_utils?

yaksh/views.py
```diff
@@ -73,13 +74,30 @@ def is_moderator(user):
...
             user.groups.add(group)
 
+
+def extract_files(questions_file):
+    if zipfile.is_zipfile(questions_file):
+        zip_file = zipfile.ZipFile(questions_file, 'r')
+        zip_file.extractall()
+
+
+def read_json(json_file, user):
```
prathamesh920: This is related to Question model then should be in models either as model method or in Question manager.

yaksh/models.py
```diff
@@ -278,6 +288,41 @@ def get_test_case(self, **kwargs):
...
+            zip_file.write(f.file.path, (os.path.basename(f.file.path)))
+            files_list.append(((os.path.basename(f.file.path)), f.extract))
+        return files_list
+
+    def _add_files_to_db(self, file_names):
+        for file_name, extract in file_names:
+            que_file = open(file_name, 'r')
+            #Converting to Python file object with some Django-specific additions
+            django_file = File(que_file)
+            f = FileUpload.objects.get_or_create(file=django_file,
```
prathamesh920: Why do you need f?

yaksh/management/commands/create_demo_course.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def create_demo_course():
+    """ creates a demo course, quiz """
+
+    success = False
+    print("Creating Demo User...")
+    # create a demo user
+    user, u_status = User.objects.get_or_create(username='demo_user',
+                                                password='demo',
```
prathamesh920: If you create password as plain text then you won't be able to login via interface because django converts the plain password to hash. You can use set_password method.
If you are not going to use this user via interface then its fine.

yaksh/management/commands/create_demo_course.py
```diff
@@ -0,0 +1,77 @@
...
+
+    print("Creating Demo Course...")
+    # create a demo course
+    course, c_status = Course.objects.get_or_create(name="Demo_course",
+                                                    enrollment="open",
+                                                    creator=user)
+
+    print("Creating Demo Quiz...")
+    # create a demo quiz
+    quiz, q_status = Quiz.objects.get_or_create(start_date_time=timezone.now(),
```
prathamesh920: If I run the command again then again it will create a new quiz followed by new question paper, since timezone.now() will change each time.

prathamesh920:
I saw the yaksh/fixtures/demo_questions.zip,  in which there is a json file that loads the question.
In the json question **summary is "Demo Question"** for all questions.

And the create_demo_course command there is a following line:
questions = Question.objects.filter(active=True, **summary="Demo_Question"**)

These questions are added to the question paper, but the above query will return empty queryset as the summary won't match and no questions will be added to the fixed questions!
So the question paper will be created without any questions.

Kindly check and test the interface whenever there is a new commit.

prabhuramachandran:
@adityacp -- please address the issues raised by @prathamesh920 ASAP.

adityacp:
@prathamesh920 I have added condition to check if questions with summary as `Demo Question` exists. If not, it will print error message.

yaksh/bash_files/sample2.args
```diff
@@ -0,0 +1 @@
+file1.csv file2.csv file3.csv
```
prabhuramachandran: I do not like the fact that this is being checked in at all on the server.

prabhuramachandran: For now we can merge this but I want this cleaned up ASAP.

yaksh/views.py
```diff
@@ -252,8 +256,9 @@ def add_quiz(request, course_id, quiz_id=None):
     user = request.user
     course = get_object_or_404(Course, pk=course_id)
     ci = RequestContext(request)
-    if not is_moderator(user) or (user != course.creator and user not in course.teachers.all()):
-        raise Http404('You are not allowed to view this page!')
+    if course.name != "Yaksh_Demo_course":
```
prabhuramachandran: This is rather ugly.

yaksh/management/commands/create_demo_course.py
```diff
@@ -0,0 +1,92 @@
...
+    Profile.objects.get_or_create(user=user, roll_number=0,
+                                  institute='demo_institute',
+                                  department='demo_department',
+                                  position='Faculty')
+    demo_user = User.objects.filter(username='yaksh_demo_user')
+    add_to_group(demo_user)
+
+    print("Creating Demo Course...")
+    # create a demo course
+    course, c_status = Course.objects.get_or_create(name="Yaksh_Demo_course",
```
prabhuramachandran: Will this look ugly on the UI?

yaksh/test_views.py
```diff
@@ -650,7 +650,9 @@ def setUp(self):
         self.client = Client()
 
         self.mod_group = Group.objects.create(name='moderator')
-
+        User.objects.get_or_create(username='demo_user',
```
adityacp: Resolved

yaksh/views.py
adityacp: Resolved

yaksh/management/commands/create_demo_course.py
adityacp: Resolved. Changed name from `Yaksh_Demo_course` to `Yaksh Demo course`

yaksh/models.py
```diff
@@ -150,6 +156,23 @@ def get_teachers(self):
     def remove_teachers(self, *teachers):
         self.teachers.remove(*teachers)
 
+    def create_demo(self, user):
+        course = Course.objects.filter(creator=user, name="Yaksh Demo course")
+        if not course:
+            course, c_status= Course.objects.get_or_create(name="Yaksh Demo course",
```
ankitjavalkar: Since you are already checking if a course exists, just do a "create" over here.

ankitjavalkar:
Looks good

yaksh/models.py
```diff
@@ -150,6 +156,23 @@ def get_teachers(self):
...
+        if not course:
+            course = Course.objects.create(name="Yaksh Demo course",
+                                           enrollment="open",
+                                           creator=user)
+            quiz = Quiz()
+            demo_quiz = quiz.create_demo_quiz(course)
+            demo_ques = Question()
+            demo_ques.create_demo_questions(user)
+            demo_que_ppr = QuestionPaper()
+            demo_que_ppr.create_demo_que_ppr(demo_quiz)
```
prabhuramachandran: Can this method be `create_demo_quiz_paper`?

yaksh/models.py
```diff
@@ -573,6 +654,17 @@ def is_prerequisite_passed(self, user):
         if self.quiz.has_prerequisite():
             prerequisite = self._get_prequisite_paper()
             return prerequisite._is_questionpaper_passed(user)
+
+    def create_demo_que_ppr(self, demo_quiz):
```
prabhuramachandran: Please change this to `create_demo_quiz_paper`.

yaksh/templates/manage.html
```diff
@@ -77,6 +77,26 @@
...
+                    <button class="btn" type="button" onClick='location.replace("{{URL_ROOT}}/exam/manage/create_demo_course");'>Create Demo Course</button>
+                    <div style="display: none;" id="help">
+                    <ol>
+                    <ul>
+                    <li>A Demo Course and Demo Quiz will be created (Click Courses link on nav bar to view courses).</li>
+                    <li>Some Demo Questions are also created for you (Click Questions link on nav bar to view questions).</li>
+                    <li>In Courses you can view Demo Quiz.</li>
+                    <li>Click on the Demo Quiz and Click on User Mode or God Mode to take the quiz.
+                    </li>
+                    <li>You can also edit Demo quiz.
```
prabhuramachandran: `edit Demo` to `edit the Demo`

yaksh/test_models.py
```diff
@@ -58,7 +62,11 @@ def tearDownModule():
     User.objects.all().delete()
     Question.objects.all().delete()
     Quiz.objects.all().delete()
-
+    
+    que_id_list = ["25", "22", "24", "27"]
```
prathamesh920: You may simply keep the name question_ids. _list not required.

PR #128, opened 2016-08-29
Bug fixes

prathamesh920:
Issues:
1. django views taking cent percent CPU - resolved: used list instead of cycle
2. Question navigator - resolved
3. Answered saved multiple times - resolved: checks before saving whether question already answered or not.
4. Marks obtained shows zero even when answer is correct: resolved
5. Monitor page shows both skipped and submitted - resolved: shows only submitted
6. Monitor page shows incorrect answer for correct mcq and mcc - resolved

Have added test cases for next_question(skip) method of AnswerPaper.

### DIFF
--- a/yaksh/views.py
+++ b/yaksh/views.py
@@ -460,10 +460,12 @@ def skip(request, q_id, next_q=None, attempt_num=None, questionpaper_id=None):
                             correct=False, skipped=True)
         new_answer.save()
         paper.answers.add(new_answer)
-    if next_q is None:
-        next_q = paper.skip(q_id) if paper.skip(q_id) else question
-    else:
+    if next_q is not None:
         next_q = get_object_or_404(Question, pk=next_q)
+        if next_q not in paper.questions_unanswered.all():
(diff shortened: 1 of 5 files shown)


yaksh/views.py
```diff
@@ -454,16 +454,22 @@ def skip(request, q_id, next_q=None, attempt_num=None, questionpaper_id=None):
...
     if request.method == 'POST' and question.type == 'code':
         user_code = request.POST.get('answer')
         new_answer = Answer(question=question, answer=user_code,
                             correct=False, skipped=True)
         new_answer.save()
         paper.answers.add(new_answer)
-    if next_q is None:
-        next_q = paper.skip(q_id) if paper.skip(q_id) else question
-    else:
+    if next_q is not None:
```
prabhuramachandran: Is this correct, you have inverted the condition?

prathamesh920: Yes Sir, because previously it was if..else, now it is if not..else, and the statement that was in else I have written in if and vice versa.

PR #131, opened 2016-09-01
Fix quiz templates

adityacp:
Changed Quiz  interface View for students

Made tracebacks proper with line number and output text area is added below program text area.
quiz_output_traceback

Changed  `Language: Python` to `Python Code Question`
quiz_view

### DIFF
--- a/yaksh/code_evaluator.py
+++ b/yaksh/code_evaluator.py
@@ -108,9 +108,14 @@ class CodeEvaluator(object):
 
         except TimeoutException:
             err = self.timeout_msg
+        except OSError:
+            msg = traceback.format_exc(limit=0)
+            err = "Error: {0}".format(msg)
         except Exception:
-            err = "Error: {0}".format(traceback.format_exc(limit=0))
-
+            exc_type, exc_value, exc_tb = sys.exc_info()
+            tb_list = traceback.format_exception(exc_type, exc_value, exc_tb)
+            del tb_list[1:3]
+            err = "Error: {0}".format("".join(tb_list))
         finally:
             # Set back any original signal handler.
             set_original_signal_handler(prev_handler)
--- a/yaksh/code_evaluator.py
+++ b/yaksh/code_evaluator.py
@@ -114,7 +114,10 @@ class CodeEvaluator(object):
         except Exception:
             exc_type, exc_value, exc_tb = sys.exc_info()
             tb_list = traceback.format_exception(exc_type, exc_value, exc_tb)
-            del tb_list[1:3]
+            if len(tb_list) > 2:
+                del tb_list[1:3]
+            else:
+                pass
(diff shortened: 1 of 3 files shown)


yaksh/evaluator_tests/test_python_evaluation.py
```diff
@@ -139,7 +139,7 @@ def add(a, b):
         result = get_class.evaluate(**kwargs)
         err = result.get("error").splitlines()
         self.assertFalse(result.get("success"))
-        self.assertEqual(2, len(err))
+        self.assertEqual(969, len(err))
```
prathamesh920: Will always be 969?

adityacp: for that particular test case example it is fixed

ankitjavalkar:
Code is mergeable

yaksh/code_evaluator.py
```diff
@@ -108,9 +108,14 @@ def safe_evaluate(self, user_answer, test_case_data, file_paths=None):
...
             err = self.timeout_msg
+        except OSError:
+            msg = traceback.format_exc(limit=0)
+            err = "Error: {0}".format(msg)
         except Exception:
-            err = "Error: {0}".format(traceback.format_exc(limit=0))
-
+            exc_type, exc_value, exc_tb = sys.exc_info()
+            tb_list = traceback.format_exception(exc_type, exc_value, exc_tb)
+            del tb_list[1:3]
```
prabhuramachandran: I am not sure about this.  I don't think it is a good idea to do this and indiscriminately remove the two lines from `tb_list`.  What if tb_list has only two lines in it?  Can you pease ensure that this is also fully tested for different cases or write it a bit more safely.

yaksh/templates/yaksh/question.html
```diff
@@ -176,10 +171,20 @@ <h3 class="brand"><strong>Online Test</h3></strong>
...
 
          <h4>Program:</h4>
-        <textarea name="answer" id="answer"></textarea> 
-<br>
+         <textarea name="answer" id="answer"></textarea>
+         <br>
 
          {% endif %}
+         {% if question.type == "code" %}
+         <br><h4>Output:</h4></br>
```
prabhuramachandran: <br/> is all that is needed, br does not need a /br like here.

yaksh/code_evaluator.py
adityacp: I have added test cases for some exceptions and I have manually tested almost all built-in exceptions from interface also. It worked fine. 
lines which are removed from `tb_list` contains file path from which error occurred which can be removed.

prabhuramachandran: At least check if len(tb_list) >2 before deleting the lines?

PR #134, opened 2016-09-06
Selenium test with qdemo

ankitjavalkar:
Please merge only **after** merging #127


yaksh/bash_files/sample2.args
```diff
@@ -0,0 +1 @@
+file1.csv file2.csv file3.csv
```
prabhuramachandran: Why are these needed?

yaksh/bash_files/sample2.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+cat $1 | cut -d: -f2 | paste -d: $3 - $2
```
prabhuramachandran: Why is this needed?  I thought the file upload allowed us to do this nicely?

yaksh/tests/load_test.py
```diff
@@ -0,0 +1,60 @@
...
+        code_server_pool = ServerPool(ports=settings.SERVER_PORTS, pool_port=settings.SERVER_POOL_PORT)
+        cls.code_server_pool = code_server_pool
+        cls.code_server_thread = t = Thread(target=code_server_pool.run)
+        t.start()
+
+        demo_course_setup = create_demo_course()
+
+    @classmethod
+    def tearDownClass(cls):
+        # User.objects.all().delete()
```
prabhuramachandran: Why is there commented code here?

.travis.yml
```diff
@@ -14,11 +14,16 @@ install:
...
   - sudo apt-get update -qq
   - sudo apt-get install -y scilab
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
+  - sleep 3 # give xvfb some time to start
 
 # command to run tests
 script:
   - python manage.py test -v 2 yaksh
+  - python manage.py test -v 2 yaksh.tests.load_test
```
prabhuramachandran: Why is this not run along with the rest?  Given the name of the class, won't it automatically be picked up?  At least a comment at a relevant location for this would be important.

yaksh/bash_files/sample2.sh
adityacp: This sample script is used in evaluator to check user answer. 
This script acts as test case for bash question.
It is not related to file uploads.
same for `sample2.args`

yaksh/tests/load_test.py
ankitjavalkar: Removed

.travis.yml
ankitjavalkar: The load_test.py uses a different type of TestCase class called LiveServerTestCase. The unit tests use the usual TestCase class.
The LiveServerTestCase class deals with Tear down a little differently - It only removes the rows of the table and does not reset the Primary keys. Because of this the subsequent model test cases fail.
Everything runs fine if both model test cases and selenium load test are run seperately.

The situation that I faced has been described in the question below. I tried a few solution but nothing seems to fix the issue
http://stackoverflow.com/questions/32998166/with-py-test-database-is-not-reset-after-liveservertestcase

I have posted this to the django user google group/mailing list, but the message seems to be pending approval.

PR #138, opened 2016-09-16
Python assertion bug fix

maheshgudi:
Previously, if a question had a correct test case followed next by an invalid test case (i.e. executing the test case would raise an Non Assertion based exception), the user would pass despite giving the wrong answer.

This bug has been fixed.

### DIFF
--- a/yaksh/code_evaluator.py
+++ b/yaksh/code_evaluator.py
@@ -101,6 +101,7 @@ class CodeEvaluator(object):
         # Do whatever testing needed.
         try:
             for test_case in test_case_data:
+                success = False
                 self.compile_code(user_answer, file_paths, **test_case)
                 success, err = self.check_code(user_answer, file_paths, **test_case)
                 if not success:
@@ -116,8 +117,6 @@ class CodeEvaluator(object):
             tb_list = traceback.format_exception(exc_type, exc_value, exc_tb)
             if len(tb_list) > 2:
                 del tb_list[1:3]
-            else:
-                pass
             err = "Error: {0}".format("".join(tb_list))
         finally:
             # Set back any original signal handler.
--- a/yaksh/evaluator_tests/test_python_evaluation.py
+++ b/yaksh/evaluator_tests/test_python_evaluation.py
@@ -52,7 +52,7 @@ class PythonAssertionEvaluationTestCases(unittest.TestCase):
         self.assertFalse(result.get('success'))
         self.assertEqual(result.get('error'), self.timeout_msg)
 
-
+    def test_syntax_error(self):
         user_answer = dedent("""
         def add(a, b);
             return a + b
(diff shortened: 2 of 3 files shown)


yaksh/code_evaluator.py
```diff
@@ -116,8 +117,6 @@ def safe_evaluate(self, user_answer, test_case_data, file_paths=None):
             tb_list = traceback.format_exception(exc_type, exc_value, exc_tb)
             if len(tb_list) > 2:
                 del tb_list[1:3]
```
prabhuramachandran: Why is this block removed?

yaksh/evaluator_tests/test_python_evaluation.py
```diff
@@ -207,6 +207,57 @@ def ans():
...
+                          ]
+        syntax_error_msg = ["Traceback",
+                          "call",
+                          "File",
+                          "line",
+                          "<string>",
+                          "SyntaxError",
+                          "invalid syntax"
+                          ]
+        get_class = PythonAssertionEvaluator()
```
prabhuramachandran: Why is this called `get_class`, isn't it the evaluator?  `evaluator` should be a better name.

yaksh/evaluator_tests/test_python_evaluation.py
```diff
@@ -207,6 +207,57 @@ def ans():
         self.assertEqual(result.get('error'), "Correct answer")
         self.assertTrue(result.get('success'))
 
+    def test_single_testcase_error(self):
+        user_answer = "def palindrome(a):\n\treturn a == a[::-1]"
+        test_case_data = [{"test_case": 's="abbb"\nasert palindrome(s)==False'}
+                          ]
+        syntax_error_msg = ["Traceback",
```
prabhuramachandran: Move this to where the error is being checked to make it easier to read.  Ideally mark "# Given, # When # Then" blocks.

yaksh/evaluator_tests/test_python_evaluation.py
```diff
@@ -207,6 +207,57 @@ def ans():
...
+        result = get_class.evaluate(**kwargs)
+        err = result.get("error").splitlines()
+        self.assertFalse(result.get("success"))
+        self.assertEqual(5, len(err))
+        for msg in syntax_error_msg:
+            self.assertIn(msg, result.get("error"))
+
+
+    def test_multiple_testcase_error(self):
+        user_answer = "def palindrome(a):\n\treturn a == a[::-1]"
```
prabhuramachandran: What is the intent of this test?  What does this test do that the previous one did not?

yaksh/python_assertion_evaluator.py
```diff
@@ -49,6 +49,8 @@ def check_code(self, user_answer, file_paths, test_case):
             err = "{0} {1} in: {2}".format(type.__name__, str(value), text)
         except TimeoutException:
             raise
+        except Exception:
```
prabhuramachandran: In theory couldn't you remove both this and the previous except block as the caller will anyway catch and report the exception?

yaksh/code_evaluator.py
maheshgudi: the else case in the if-else block did nothing but pass. Hence removing it felt reasonable.

yaksh/evaluator_tests/test_python_evaluation.py
maheshgudi: Resolved

yaksh/evaluator_tests/test_python_evaluation.py
maheshgudi: Resolved

yaksh/evaluator_tests/test_python_evaluation.py
maheshgudi: The first test is suppose to test the user answer with only one wrong test case, just to see if the exception in the test case is caught or not. 
This test case will actually test the user answer with a correct test case first and then with a wrong test case, to check if the exception in the subsequent testcases is caught or if the exception goes unnoticed.

yaksh/python_assertion_evaluator.py
maheshgudi: Yes. I have removed the Timeout Exception but I decided to catch and raise all other exceptions (except AssertionError).

maheshgudi:
Resolved all issues.

I made some PEP8 changes in the python test script without entirely realizing how huge and unnecessary diff it will create. Sorry for the slip up. Will create a separate PR in the future for such changes.

yaksh/evaluator_tests/test_python_evaluation.py
prabhuramachandran: Can you change the name of the test or add a comment to the effect.

prabhuramachandran: I see that this is done.

PR #140, opened 2016-09-23
Regrade papers

prathamesh920:
A question or whole paper can be regraded.
We can regrade as per user answer paper.

**In Next PR**
I will be adding more testcases and will clean up the code further.
Also, will make the UI much better.


yaksh/models.py
```diff
@@ -14,10 +14,17 @@
...
+from os.path import join, abspath, dirname, exists
 import shutil
 import zipfile
 import tempfile
 from file_utils import extract_files
+from yaksh.xmlrpc_clients import code_server
+
+
+# The directory where user data can be saved.
+OUTPUT_DIR = abspath(join(dirname(__file__), 'output'))
```
prabhuramachandran: This is not good, what if yaksh is installed in a non-writeable location.

yaksh/models.py
```diff
@@ -994,6 +1013,71 @@ def get_previous_answers(self, question):
         if question.type == 'code':
             return self.answers.filter(question=question).order_by('-id')
 
+    def validate_answer(self, user_answer, question, json_data=None):
```
ankitjavalkar: I guess, the "question" argument is not required.

ankitjavalkar:
Rest of the code looks good

yaksh/models.py
prathamesh920: resolved

PR #145, opened 2016-10-04
Stdio bugfixes

maheshgudi:
Minor bug fixes for -
1.  In both StdIO and assertion evaluators, whenever an infinite code was executed, the parent process (the terminal in this case) was killed but it spawned a child processes (the infinite code) which the code evaluator wouldn't kill/terminate. Fixed that bug.
2.  In view_answerpaper, grade user and user data, the bootstrap panels (the answer panels which indicated correct or incorrect answer) was decided using "correct answer" as a check. Changed it to  just check True or False conditions.
3.  Added no. of complete and inprogress papers for a quiz in monitor.

### DIFF
--- a/yaksh/evaluator_tests/test_scilab_evaluation.py
+++ b/yaksh/evaluator_tests/test_scilab_evaluation.py
@@ -7,6 +7,8 @@ import tempfile
 from yaksh import code_evaluator as evaluator
 from yaksh.scilab_code_evaluator import ScilabCodeEvaluator
 from yaksh.settings import SERVER_TIMEOUT
+import subprocess
+import time
 
 class ScilabEvaluationTestCases(unittest.TestCase):
     def setUp(self):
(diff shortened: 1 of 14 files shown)


prabhuramachandran:
Good catch, please add a test case for this.

yaksh/evaluator_tests/test_scilab_evaluation.py
```diff
@@ -71,6 +73,8 @@ def test_infinite_loop(self):
         result = get_class.evaluate(**kwargs) 
         self.assertFalse(result.get("success"))
         self.assertEqual(result.get("error"), self.timeout_msg)
+        time.sleep(4)
+        self.assertEqual(1, subprocess.call(["pgrep", "scilab-cli-bin"]))
```
prathamesh920: The over all code looks good, but I am unsure about the test cases.
pgrep a process may lead to unexpected results. In case scilab-cli-bin is run by a different process thread then the test case will fail.

prabhuramachandran: I agree with this assessment. It would be nice to have this tested differently.  A possibility is to use mock or to use psutil depending on what it is you are really intending to test.  What is the intent of this test?

maheshgudi:
Closing the PR for now. Will refactor and send a new PR.

PR #147, opened 2016-10-04
Hide questions instead of deletion

ankitjavalkar:


prathamesh920:
Looks good to me.
**`for q in questions:`** could be **`for question in questions:`**
But it's fine.

ankitjavalkar:
> Looks good to me.
for q in questions: could be for question in questions:

This has been fixed

PR #148, opened 2016-10-17
Fix extract files

adityacp:
Upload Questions as zip was extracted in current directory.
Now instead of extracting in current directory zip will be extracted in temp directory to avoid permissions issues.


ankitjavalkar:
Looks good

PR #152, opened 2016-11-07
Partial Grading

ankitjavalkar:
Add partial grading to evaluation

### DIFF
--- a/yaksh/bash_code_evaluator.py
+++ b/yaksh/bash_code_evaluator.py
@@ -49,9 +49,11 @@ class BashCodeEvaluator(CodeEvaluator):
 
         Returns (False, error_msg): If mandatory arguments are not files or if
         the required permissions are not given to the file(s).
-
         """
         ref_code_path = test_case
+        success = False
+        test_case_weightage = 0.0
+
         get_ref_path, get_test_case_path = ref_code_path.strip().split(',')
         get_ref_path = get_ref_path.strip()
         get_test_case_path = get_test_case_path.strip()
@@ -62,19 +64,16 @@ class BashCodeEvaluator(CodeEvaluator):
             self.files = copy_files(file_paths)
         if not isfile(clean_ref_code_path):
             msg = "No file at %s or Incorrect path" % clean_ref_code_path
-            return False, msg
+            return False, msg, 0.0
         if not isfile(self.submit_code_path):
             msg = "No file at %s or Incorrect path" % self.submit_code_path
-            return False, msg
+            return False, msg, 0.0
         if not os.access(clean_ref_code_path, os.X_OK):
             msg = "Script %s is not executable" % clean_ref_code_path
-            return False, msg
+            return False, msg, 0.0
         if not os.access(self.submit_code_path, os.X_OK):
             msg = "Script %s is not executable" % self.submit_code_path
-            return False, msg
-
-        success = False
-        test_case_weightage = 0.0
+            return False, msg, 0.0
 
         user_answer = user_answer.replace("\r", "")
         self.write_to_submit_code_file(self.submit_code_path, user_answer)
@@ -99,14 +98,14 @@ class BashCodeEvaluator(CodeEvaluator):
                 err = "Error: expected %s, got %s" % (inst_stderr,
                     stdnt_stderr
                 )
-                return False, err, test_case_weightage
+                return False, err, 0.0
         else:
             if not isfile(clean_test_case_path):
                 msg = "No test case at %s" % clean_test_case_path
-                return False, msg, test_case_weightage
+                return False, msg, 0.0
             if not os.access(clean_ref_code_path, os.R_OK):
                 msg = "Test script %s, not readable" % clean_test_case_path
-                return False, msg, test_case_weightage
+                return False, msg, 0.0
             # valid_answer is True, so that we can stop once a test case fails
             valid_answer = True
             # loop_count has to be greater than or equal to one.
@@ -143,4 +142,4 @@ class BashCodeEvaluator(CodeEvaluator):
                     " {0}, got {1}").format(inst_stdout+inst_stderr,
                         stdnt_stdout+stdnt_stderr
                     )
-                return False, err, test_case_weightage
+                return False, err, 0.0
--- a/yaksh/bash_code_evaluator.py
+++ b/yaksh/bash_code_evaluator.py
@@ -28,7 +28,7 @@ class BashCodeEvaluator(CodeEvaluator):
             delete_files(self.files)
         super(BashCodeEvaluator, self).teardown()
 
-    def check_code(self, user_answer, file_paths, partial_grading, test_case, weightage):
+    def check_code(self, user_answer, file_paths, partial_grading, test_case, weight):
         """ Function validates student script using instructor script as
         reference. Test cases can optionally be provided.  The first argument
         ref_path, is the path to instructor script, it is assumed to
@@ -52,7 +52,7 @@ class BashCodeEvaluator(CodeEvaluator):
         """
         ref_code_path = test_case
         success = False
-        test_case_weightage = 0.0
+        test_case_weight = 0.0
 
         get_ref_path, get_test_case_path = ref_code_path.strip().split(',')
         get_ref_path = get_ref_path.strip()
@@ -92,8 +92,8 @@ class BashCodeEvaluator(CodeEvaluator):
             )
             proc, stdnt_stdout, stdnt_stderr = ret
             if inst_stdout == stdnt_stdout:
-                test_case_weightage = float(weightage) if partial_grading else 0.0
-                return True, "Correct answer", test_case_weightage
+                test_case_weight = float(weight) if partial_grading else 0.0
+                return True, "Correct answer", test_case_weight
             else:
                 err = "Error: expected %s, got %s" % (inst_stderr,
                     stdnt_stderr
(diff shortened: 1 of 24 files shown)


yaksh/bash_code_evaluator.py
```diff
@@ -27,7 +28,7 @@ def teardown(self):
             delete_files(self.files)
         super(BashCodeEvaluator, self).teardown()
 
-    def check_code(self, user_answer, file_paths, test_case):
+    def check_code(self, user_answer, file_paths, partial_grading, test_case, marks):
```
prathamesh920: check_code now returns 3 values, so it will be better if the Returns description is modified in the method docstring.

yaksh/models.py
```diff
@@ -1104,9 +1115,11 @@ class TestCase(models.Model):
 
 class StandardTestCase(TestCase):
     test_case = models.TextField(blank=True)
+    marks = models.FloatField(default=0.0)
```
prathamesh920: So each test case will have marks!
Instead I think a single constant fraction value for each testcase is a better option.
Else we can sum the total testcases and find the ratio of right to total.

yaksh/models.py
```diff
@@ -1077,11 +1083,16 @@ def regrade(self, question_id):
         json_data = question.consolidate_answer_data(answer) \
                             if question.type == 'code' else None
         correct, result = self.validate_answer(answer, question, json_data)
-        user_answer.marks = question.points if correct else 0.0
         user_answer.correct = correct
         user_answer.error = result.get('error')
+        if correct:
```
prathamesh920: If we have a fraction(value 0 to 1) returned from the code sever, then we do not need  this if block, we can simply write 
```user_answer.marks = question.points * result['marks']```
if wrong then marks is 0
if right then marks is 1
else fraction of marks.
So no need to check correct or wrong for partial grading when adding marks.

Comment (ankitjavalkar) [inline yaksh/models.py]: If condition will have to stay.
In case when there is no partial grading for a question, in that case the evaluator always returns 0.0
So if a question is answered correctly but it is not graded partially then it will end up being marked 0.0 marks (in case success=True is not checked)

File: yaksh/bash_code_evaluator.py (hunk 1 above)
Comment (prabhuramachandran) [inline yaksh/bash_code_evaluator.py]: The docstring can be updated now though.

File: yaksh/models.py (hunk 3 above)
Comment (prabhuramachandran) [inline yaksh/models.py]: I agree with @prathamesh920 that each test case having marks is non optimal.  I think I have a nicer way to specify this.  Basically tread the "marks" you have above as a weight for each test, then the total mark is sum(obtained weight)/total(weight)*marks.  By default each test case has a weight of 1, this way is nicer I think and much easier to implement.  A teacher may even change the total marks without having to worry about the individual weights.

File: yaksh/models.py (hunk 2 above)
Comment (prabhuramachandran) [inline yaksh/models.py]: Agree with @prathamesh920 !

File: yaksh/models.py (hunk 3 above)
Comment (ankitjavalkar) [inline yaksh/models.py]: Done

**Comment by ankitjavalkar** [comment]
- Docstrings added.
- Test cases now have point based weightage

File: yaksh/views.py [hunk 4]
```diff
@@ -517,11 +517,16 @@ def check(request, q_id, attempt_num=None, questionpaper_id=None):
                         if question.type == 'code' else None
         correct, result = paper.validate_answer(user_answer, question, json_data)
         if correct:
+            new_answer.marks = (question.points * result['weightage'] /
```
prathamesh920: in code_evaluator line number 92 is following
```result = {'success': success, 'error': error, 'marks': marks}```
so here ``` result['weightage']``` will throw an error!

File: yaksh/views.py [hunk 5]
```diff
@@ -517,11 +517,16 @@ def check(request, q_id, attempt_num=None, questionpaper_id=None):
...
-            new_answer.marks = question.points
             new_answer.error = result.get('error')
         else:
             new_answer.error = result.get('error')
+            new_answer.marks = (question.points * result['weightage'] /
+                question.get_maximum_test_case_weightage()) \
+                if question.partial_grading and question.type == 'code' else 0
         new_answer.save()
         paper.update_marks('inprogress')
         paper.set_end_time(timezone.now())
```
prathamesh920: Below we see that there is check ```if not result.get('success'):```
in else we add the current question to completed questions and show the next question.
So in case my all testcases fail except for the last one then the question will be added to the completed questions!!!

yaksh/code_evaluator.py
```diff
@@ -108,32 +108,41 @@ def safe_evaluate(self, user_answer, test_case_data, file_paths=None):
...
+                success, err, test_case_weightage = self.check_code(user_answer,
+                    file_paths,
+                    partial_grading,
+                    **test_case
+                )
+                if success:
+                    weightage += test_case_weightage
+                    error = err
+                else:
+                    error += err + "\n"
```
prathamesh920: So for all testcases we are appending the err.
In case, the first testcase passes and the next testcase fails then
**error message will be Correct Answer + Assertion Error.**

This can be covered in your next PR related to error message formatting,

prabhuramachandran: So maybe error should be a list.

yaksh/bash_code_evaluator.py
```diff
@@ -27,7 +28,7 @@ def teardown(self):
             delete_files(self.files)
         super(BashCodeEvaluator, self).teardown()
 
-    def check_code(self, user_answer, file_paths, test_case):
+    def check_code(self, user_answer, file_paths, partial_grading, test_case, weightage):
```
prabhuramachandran: Instead of `weightage` perhaps use `weight`.

yaksh/bash_code_evaluator.py
```diff
@@ -133,10 +136,11 @@ def check_code(self, user_answer, file_paths, test_case):
...
-                return True, "Correct answer"
+                test_case_weightage = float(weightage) if partial_grading else 0.0
+                return True, "Correct answer", test_case_weightage
             else:
                 err = ("Error:expected"
                     " {0}, got {1}").format(inst_stdout+inst_stderr,
                         stdnt_stdout+stdnt_stderr
                     )
-                return False, err
+                return False, err, test_case_weightage
```
prabhuramachandran: Why is this returning the weightage?

yaksh/xmlrpc_clients.py
```diff
@@ -62,7 +62,9 @@ def run_code(self, language, test_case_type, json_data, user_dir):
             server = self._get_server()
             result = server.check_code(language, test_case_type, json_data, user_dir)
         except ConnectionError:
-            result = json.dumps({'success': False, 'error': 'Unable to connect to any code servers!'})
+            result = json.dumps({'success': False,
+                'weightage': 0.0,
+                'error': 'Unable to connect to any code servers!'})
```
prabhuramachandran: Please describe what the return values are in the docstring very clearly.  This looks like the sum of the weights that the user code passed, correct?

yaksh/views.py
ankitjavalkar: I had already fixed this in a minor commit, It did not get pushed. I apologise. This has been fixed

yaksh/bash_code_evaluator.py
ankitjavalkar: I have initialised the test_case_weightage as 0.0 earlier. Hence the variable is returned as is. 

However I agree that this may be a bit unclear, hence fixing this to return a constant 0.0

yaksh/xmlrpc_clients.py
ankitjavalkar: Yes, that is correct, Fixed the docstrings accordingly

yaksh/bash_code_evaluator.py
ankitjavalkar: Fixed

yaksh/views.py
ankitjavalkar: This has been fixed now

ankitjavalkar:
Please check the Output Display Interface:

ankitjavalkar:
- Added a Test case to verify the situation pointed out by @prathamesh920

yaksh/bash_code_evaluator.py
```diff
@@ -39,41 +40,44 @@ def check_code(self, user_answer, file_paths, test_case):
...
-        returns (True, "Correct answer") : If the student script passes all
+        returns (True, "Correct answer", 1.0) : If the student script passes all
         test cases/have same output, when compared to the instructor script
 
-        returns (False, error_msg): If the student script fails a single
+        returns (False, error_msg, 0.0): If the student script fails a single
         test/have dissimilar output, when compared to the instructor script.
 
-        Returns (False, error_msg): If mandatory arguments are not files or if
+        Returns (False, error_msg, 0.0): If mandatory arguments are not files or if
```
prabhuramachandran: In this case, should this just raise a custom exception?

yaksh/python_assertion_evaluator.py
```diff
@@ -47,11 +68,13 @@ def check_code(self, user_answer, file_paths, test_case):
             info = traceback.extract_tb(tb)
             fname, lineno, func, text = info[-1]
             text = str(test_case).splitlines()[lineno-1]
-            err = "{0} {1} in: {2}".format(type.__name__, str(value), text)
+            err = ("-----\nExpected Test Case:\n{0}\n"
```
prabhuramachandran: Looks ugly, can't it be some HTML code or can this be delegated to something else?

prabhuramachandran:
Merging for now.

PR #158, opened 2016-11-18
Fix file upload and tooltip

adityacp:
Now for a question We can hide the file if it is not to be shown to students.
Fixed Tooltip rendering.
Some Minor Jquery fixes.


ankitjavalkar:
Looks Good, Mergeable.

yaksh/models.py
```diff
@@ -401,6 +402,12 @@ def set_extract_status(self):
             self.extract = True
         self.save()
 
+    def set_hide_status(self):
```
prabhuramachandran: Change this to `toggle_hide_status` or you can just pass this as an argument and set the value and save it.

PR #160, opened 2016-11-18
Date time picker and python assertion exception fix

maheshgudi:
1. Added Date time picker for setting quiz time and date.
2. Fixed python assertion exception i.e now the exception in the python assertion evaluator will return the traceback to the code evaluator as a string instead of raising it to the code evaluator exception.


yaksh/python_assertion_evaluator.py
```diff
@@ -69,9 +69,13 @@ def check_code(self, user_answer, file_paths, partial_grading, test_case, weight
...
             err = ("-----\nExpected Test Case:\n{0}\n"
-                    "Error - {1} {2} in: {3}\n-----").format(test_case, type.__name__, str(value), text)
+                    "Error - {1} {2} in: {3}\n-----").format(test_case,
+                                                             type.__name__,
+                                                             str(value), text
+                                                             )
         except Exception:
-            raise  # Exception will be caught in CodeEvaluator.
+             msg = traceback.format_exc(limit=0)
+             err = "Error in Test case: {0}".format(msg)
```
prabhuramachandran: you need to raise if timeout is raised before Exception

PR #167, opened 2016-12-23
Modify Evaluators to take moderator code as test case instead of files

adityacp:
This PR will allow the moderator to write code while creating questions instead of submitting a file path in test case field.


yaksh/models.py
```diff
@@ -1148,11 +1148,14 @@ class TestCase(models.Model):
 class StandardTestCase(TestCase):
     test_case = models.TextField()
     weight = models.FloatField(default=1.0)
+    test_case_args = models.TextField(help_text="<b>Command Line arguments for bash only</b>",
+                                      blank=True)
```
adityacp: I have added a field for command line arguments in `StandardTestCase` for bash.
I am not sure whether to add it in `StandardTestCase` or create a new Test Case.

ankitjavalkar:
Looks good, Mergeable.

prabhuramachandran:
Can you please save the questions somewhere so we can use this?

yaksh/models.py
adityacp: @prabhuramachandran Sir, could you please suggest me with the above comment?

PR #168, opened 2016-12-23
Fixed get_test_cases method

maheshgudi:
This PR contains the following fixes:
1. Previously, questions with multiple test cases of the same type would return a **MultipleObjectsReturn** Exception. Fixed the above mentioned bug.

2. Also, changed the Stdio Test cases model class attributes to be a **Textfield** instead if Charfield as it would not take newline entries in the form.


ankitjavalkar:
LGTM, on priority.

PR #176, opened 2017-01-09
Hook evaluator

maheshgudi:
The following PR is for python hook evaluator.

This evaluator takes the user answer of the student and evaluates it against the hook code provided by the moderator.


yaksh/evaluator_tests/test_java_evaluation.py
```diff
@@ -507,5 +507,334 @@ class Test{
...
+                             class Test
+                             {public static void main(String[] args){
+                             while(0==0)
+                            {
+                            System.out.print("a");}
+                            }}""")
+
+        hook_code = dedent("""\
+                            def check_answer(user_answer):
+                                with open("Test.java", "w+") as f:
```
prabhuramachandran: Do you really need such a complex test?  All you have is either the test code returning different values depending on the answer.  How that is actually evaluated is a detail.  While the test does serve as an interesting example I am afraid it may make maintaining the tests difficult.  Many of the other tests are also similar.

yaksh/evaluator_tests/test_python_evaluation.py
```diff
@@ -643,5 +641,217 @@ def test_infinite_loop(self):
...
+        
+    def test_infinite_loop(self):
+        # Given
+        user_answer = "def add(a, b):\n\twhile True:\n\t\tpass"
+        hook_code = dedent("""\
+                            def check_answer(user_answer):
+                                success = False
+                                err = "Incorrect Answer"
+                                mark_fraction = 0.0
+                                exec(user_answer, globals())
```
prathamesh920: We can exec this in check_code method of HookEvaluator. Then a moderator don't need to worry about it, can directly use the function defined by the user.

yaksh/hook_evaluator.py
```diff
@@ -0,0 +1,65 @@
...
+
+        Returns (False, error_msg, 0.0): If mandatory arguments are not files or if
+        the required permissions are not given to the file(s).
+        """
+        success = False
+        mark_fraction = 0.0
+        try:
+            tb = None
+            _tests = compile(self.hook_code, '<string>', mode='exec')
+            exec(_tests, globals())
```
prathamesh920: looks good, just a thought, instead of globals() can we use our scope dict, like
```
            d = {}
            exec(user_answer, d)
            exec(hook_code, d)
            check = d["check_answer"]
            success, err, mark_fraction = check(user_answer)
```
I think globals() is fine because moderator may also want to exec

yaksh/evaluator_tests/test_python_evaluation.py
maheshgudi: Resolved

ankitjavalkar:
Looks good

PR #177, opened 2017-01-11
Fix management commands

adityacp:
This PR will fix the yaksh management commands for a quick start.


setup.py
```diff
@@ -27,7 +27,7 @@
     name='yaksh',
     author='Python Team at FOSSEE, IIT Bombay',
     author_email='python@fossee.in',
-    version='0.1.1',
+    version='0.1.3',
```
prathamesh920: Is this  an appropriate version number?

ankitjavalkar:
LGTM

PR #178, opened 2017-01-12
Added search option in answerpaper admin page, gitignore doc _builds

maheshgudi:
1. Added search option in answerpaper admin page, helpful to find specific answerpapers.
2. Modified gitinore file, ignores _builds in docs now


ankitjavalkar:
LGTM

PR #179, opened 2017-01-12
Fixed a bug which would mark all MCQ and MCC questions' answers as correct

maheshgudi:
1. Due to the new evaluator, an unchecked bug, which marked all mcq and mcc attempts as correct, despite actually marking the wrong answer, has been fixed.
2. Added Test case to check for incorrect mcq and mcc questions
3. question.html would also render error msg for mcq and mcc questions. This has also been fixed.

### DIFF
--- a/yaksh/templates/yaksh/question.html
+++ b/yaksh/templates/yaksh/question.html
@@ -161,33 +161,11 @@ function call_skip(url)
                  </div>
                  <div class="panel-body">
                      {% if question.type == "mcq" %}
-                    {% if error_message %}
-                        <p>
-                        <div class="panel panel-danger">
-                        <div class="panel-heading">
-                        {% for err in error_message %}
-                            {{ err }}
-                        {% endfor %}
-                        </div>
-                        </div>
-                        </p>
-                     {% endif %}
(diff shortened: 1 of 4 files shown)


yaksh/test_models.py
```diff
@@ -607,6 +607,42 @@ def test_validate_and_regrade_mcq_question(self):
...
+        self.answerpaper.answers.add(self.answer)
+
+        # When
+        json_data = None
+        result = self.answerpaper.validate_answer(mcq_answer,
+                                                  self.question2, json_data
+                                                  )
+
+        # Then
+        self.assertTrue(result['success'])
```
prathamesh920: Should this be assertFalse?

yaksh/templates/yaksh/question.html
```diff
@@ -161,33 +161,11 @@
                  </div>
                  <div class="panel-body">
                      {% if question.type == "mcq" %}
```
prathamesh920: For MCQ and MCC, error message is shown only when user do not select any option and submits.
Error message is shown, "Please select a valid option"

adityacp:
Merging because of priority

PR #184, opened 2017-01-23
Minor view changes

maheshgudi:
1. Monitor answer paper view modified to accommodate the new changes in code evaluator.
2. Changed options attribute in mcqfield model method to text field instead of charfield.
3. moved activate/deactivate course button to the top of the course_details page.
4. Monitor page reloads with the table sorted wrt the marks obtained by students, in descending order.
5. Removed an unnecessary variable instructions_data from models.

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -572,7 +572,6 @@ class Quiz(models.Model):
                                         end_date_time=timezone.now() + timedelta(176590),
                                         duration=30, active=True,
                                         attempts_allowed=-1,
-                                        instructions=instructions_data,
                                         time_between_attempts=0,
                                         description='Yaksh Demo quiz', pass_criteria=0,
                                         language='Python', prerequisite=None,
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -68,43 +68,6 @@ test_status = (
                 ('completed', 'Completed'),
               )
 
-instructions_data = dedent("""\
-        <p>
-        This examination system has been developed with the intention of
-        making you learn programming and be assessed in an interactive and
-        fun manner. You will be presented with a series of programming questions
-        and problems that you will answer online and get immediate
-        feedback for.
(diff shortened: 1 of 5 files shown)


yaksh/models.py
```diff
@@ -68,43 +68,6 @@
                 ('completed', 'Completed'),
               )
 
-instructions_data = dedent("""\
```
adityacp: This variable is used in create_demo_quiz method for instructions in quiz models.
Please check if that works for creating demo course and demo quiz.

maheshgudi: This has been moved to forms.py in a previous PR. The default instructions is initialized from forms and not models. Hence it is no longer required in models.

maheshgudi: After re-reading your comment, I understand that the removal of instructions_data would have caused a problem in creating a demo quiz. Hence, i have removed the instructions argument from create_demo_quiz model method. Instead while creating the demo course, it will take the default value as defined in the instruction attribute of quiz model method.

yaksh/templates/yaksh/user_data.html
```diff
@@ -84,36 +84,37 @@
...
+            <div class="panel-heading">Error
+        
+        {% for error in answer.error_list %}
+        <div><pre><code>{{ error }}</code></pre></div>
         {% endfor %}
+        {% endif %}    
+    
     </div>
-    <div class="panel-body"><pre><code>{{ answer.answer.strip }}</code></pre></div>
+    <div class="panel-body"><pre><code>{{ answer.answer.answer.strip }}</code></pre></div>
```
prabhuramachandran: Wow, ``answer.answer.answer`` that is pretty nested?

maheshgudi: i admit this is dirty. Will change the variable names to something more appropriate in the next PR.

PR #185, opened 2017-01-23
Add course instructions

ankitjavalkar:
image

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -146,6 +146,7 @@ class Course(models.Model):
     created_on = models.DateTimeField(auto_now_add=True)
     teachers = models.ManyToManyField(User, related_name='teachers')
     is_trial = models.BooleanField(default=False)
+    instructions = models.TextField(null=True, blank=True)
     objects = CourseManager()
 
     def request(self, *users):
@@ -232,9 +233,8 @@ class Course(models.Model):
 
 ###############################################################################
 class ConcurrentUser(models.Model):
-    concurrent_user = models.OneToOneField(User, null=False)
-    session_key = models.CharField(null=False, max_length=40)
-
+    concurrent_user = models.OneToOneField(User)
+    session_key = models.CharField(max_length=40)
 
 ###############################################################################
 class Profile(models.Model):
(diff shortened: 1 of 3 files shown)


yaksh/models.py
```diff
@@ -146,6 +146,7 @@ class Course(models.Model):
     created_on = models.DateTimeField(auto_now_add=True)
     teachers = models.ManyToManyField(User, related_name='teachers')
     is_trial = models.BooleanField(default=False)
+    instructions = models.TextField(null=True, blank=True)
```
maheshgudi: please add a default value or none. This makes it easy for migrations.

yaksh/templates/yaksh/quizzes_user.html
```diff
@@ -60,6 +60,21 @@
             </table>
         {% endif %}
         </div>
+
+        <div class="row">
```
maheshgudi: add a if instructions then display in the template condition.

yaksh/templates/yaksh/quizzes_user.html
```diff
@@ -60,6 +60,21 @@
...
+        <div class="panel-group">
+          <div class="panel panel-default">
+            <div class="panel-heading">
+              <h4 class="panel-title">
+                <a data-toggle="collapse" href="#collapse{{ forloop.counter }}">Course Instructions</a>
+              </h4>
+            </div>
+            <div id="collapse{{ forloop.counter }}" class="panel-collapse collapse">
+              <div class="panel-body"> {{ course.instructions }}</div>
```
maheshgudi: could the course instructions be rendered as html.

ankitjavalkar: We can add HTML tags to render as HTML

yaksh/templates/yaksh/quizzes_user.html
ankitjavalkar: Done

yaksh/models.py
ankitjavalkar: Done, default is none and is only rendered if Instructions is a non Null value

ankitjavalkar:
Issues addressed, mergeable.

PR #187, opened 2017-01-29
Course related bugfixes

maheshgudi:
The following are the changes -

1.  Added bootstrap info glyphicons and collapse instead of css and jquery for help text.
2.  Moderator can now download Course CSV for course he/she was added to.
3. Using individual enroll and reject buttons for a student in course would add different user instead of the one intended; fixed this bug.
4. Added a <b>select all</b> checkbox to select all trial papers at the same time.
5. Added student info to bootstrap modal instead of tooltip in course detail page.

ezgif com-optimize

6. Added search from user ip feature for answerpaper in admin.


yaksh/templates/yaksh/course_detail.html
```diff
@@ -49,7 +88,46 @@
...
+                    <div class="modal-dialog">
+                    <div class="modal-content">
+                    <div class="modal-header">
+                        <button type="button" class="close" data-dismiss="modal">&times;</button>
+                        <h4 class="modal-title">Information for {{ rejected.username }} </h4>
+                    </div>
+                    <div class="modal-body">
+                        <table>
+                            <tr>
+                                <td><strong>Full Name :</strong></td>
```
prabhuramachandran: This modal block is repeated above and here, can this not be made common somewhere?

yaksh/templates/yaksh/course_detail.html
```diff
@@ -75,7 +153,46 @@
...
+                    <div class="modal-dialog">
+                    <div class="modal-content">
+                    <div class="modal-header">
+                        <button type="button" class="close" data-dismiss="modal">&times;</button>
+                        <h4 class="modal-title">Information for {{ enrolled.username }} </h4>
+                    </div>
+                    <div class="modal-body">
+                        <table>
+                            <tr>
+                                <td><strong>Full Name :</strong></td>
```
prabhuramachandran: Again, repeated with slightly different template variables.

maheshgudi: We can add a js dom function which will create the modal block with different attributes as and when called for.

prabhuramachandran:
Merging for now.

PR #198, opened 2017-02-01
Forgot Password and Reset Issue Resolved

prathamesh920:
The problem was:
We have included yaksh urls to the project urls with the namespace
"yaksh".
So whenever we call the url name i.e "yaksh:<url>" a reverse match is
made.
But for Forgot Password and Change Password we are using django's
in-built functionality, where the reverse url is hardcoded.
So the reverse match fails as the namespace is not specified!
To resolve this, created a urls_password_reset URL dispatcher file and
included to the project urls.


prathamesh920:
As we are using django's built-in functions, we don't need to write test for them. Instead, we can have a crawler.

prabhuramachandran:
Well in that case it should have worked from the start right?  It didn't, so having this caught by a test suite would be useful.

prabhuramachandran:
See #200.

PR #203, opened 2017-02-02
Fix #201.

### LINKED ISSUE #201: Running the code server on OSX as nobody causes problems
On a simple test demo instance, if you run `sudo yaksh run_code_server`, it is unable to create the temporary directory or so leading to an exception like so:

```
 File "envs/yaksh/lib/python3.5/xmlrpc/client.py", line 1322, in parse_response
    return u.close()
  File "envs/yaksh/lib/python3.5/xmlrpc/client.py", line 655, in close
    raise Fault(**self._stack[0])
xmlrpc.client.Fault: <Fault 1: "<class 'PermissionError'>:[Errno 13] Permission denied: '/var/folders/tr/pfk3lcgd4dndrbppzjwp_tp40000gn/T/output'">
```

prabhuramachandran:
The issue is that when the ouput directory is set to
`tempfile.gettempdir()` the user nobody cannot stat the created
directory on OS X.  The code server therefore is unable to see the
directory resulting in problems.  Instead we create the directory inside
the application as it was done earlier.

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -26,7 +26,6 @@ from yaksh.xmlrpc_clients import code_server
 from django.conf import settings
 
 
-
 languages = (
         ("python", "Python"),
         ("bash", "Bash"),
@@ -35,7 +34,6 @@ languages = (
         ("java", "Java Language"),
         ("scilab", "Scilab"),
     )
-
 
 question_types = (
         ("mcq", "Multiple Choice"),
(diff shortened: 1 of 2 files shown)


yaksh/models.py
```diff
@@ -68,21 +63,29 @@
                 ('completed', 'Completed'),
               )
 
+
 def get_assignment_dir(instance, filename):
-    return '%s/%s/%s' % (instance.user.user, instance.assignmentQuestion.id, filename)
+    return os.sep.join((
+        instance.user.user, instance.assignmentQuestion.id, filename
```
prabhuramachandran: This was already there and I did not change this, I just changed the implementation a bit.  We can change this later if needed.

maheshgudi: I know it was this way before. I commented on it because I had a painful experience of finding the assignment files. The problem is these question IDs are not displayed anywhere on the interface except for the URLs. Maybe as a quick fix we can display the question IDs on the interface somewhere. Anyways this is mergable.

PR #204, opened 2017-02-02
Cleanup yaksh script: fixes #195.

### LINKED ISSUE #195: Improve yaksh script
Currently it seems you can only have one demo instance to play with.  Instead of writing the NAME and PATH to the project_details.py it should either write a config file, or expect an additional argument to the path which should be used as the directory with the NAME as the basename of the dir.

The cli.py also does not satisfy PEP 8 and has other issues such as:

1. It does not pass any additional command line args to run_code_server.
2. The demo_settings.py seem to use an older django version.  This should be updated.

prabhuramachandran:
- `yaksh create_demo` now just takes a path and there is no project name.
- `yaksh run_demo` is now `yaksh run path_to_demo`.
- This eliminates the need for the annoying `project_detail.py`, which  is now removed.
- Many demo projects can be created and run.
- Update the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Yaksh
 [![Build Status](https://travis-ci.org/FOSSEE/online_test.svg?branch=master)](https://travis-ci.org/FOSSEE/online_test)
 [![Documentation Status](https://readthedocs.org/projects/yaksh/badge/?version=latest)](http://yaksh.readthedocs.io/en/latest/?badge=latest)
 
-To get a hang of the Yaksh interface please refer to the user documentation at [Yaksh Docs] (http://yaksh.readthedocs.io)
+To get an overview of the Yaksh interface please refer to the user documentation at [Yaksh Docs] (http://yaksh.readthedocs.io)
 
 This is a Quickstart guide to help users setup a trial instance. If you wish to deploy Yaksh in a production environment here is a [Production Deployment Guide] (https://github.com/FOSSEE/online_test/blob/master/README_production.md)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -29,41 +29,62 @@ programming quiz.
...
+#### Short instructions
+
+To see a quick demo after installing yaksh do the following:
+
+    $ yaksh create_demo yaksh_demo
+    $ yaksh run yaksh_demo
+
+On another terminal start up the code server that executes the user code safely:
+
+    $ sudo yaksh run_code_server
```
maheshgudi: At the risk of stating the obvious,  yaksh has to be installed system wide as well. It would be better to explicitly mention this here.

yaksh/scripts/cli.py
```diff
@@ -21,61 +18,68 @@
...
+
     # create parser for the "run_demo" subcommand
-    run_demo_parser = subparser.add_parser("run_demo",
-                        help="Initialise django server and run the demo project")
+    run_parser = subparser.add_parser(
+        "run",
+        help="Initialise django server and run the demo project"
+    )
+    run_parser.add_argument(
+        "path", type=str, nargs=1,
```
maheshgudi: shouldn't this have an additional argument for ports? In case if something has already occupied the default port.

README.md
prabhuramachandran: Not really, that is an issue with sudo picking up the PATH of the user or not.

yaksh/scripts/cli.py
prabhuramachandran: Can be added later.  For now this PR fixes the annoying changes to `project_details.py`.

PR #206, opened 2017-02-03
added skip question feature. closes #191

### LINKED ISSUE #191: Add option allow_skip to quiz
This should allow an instructor to create a quiz where users cannot skip a question and can proceed to the next question only when they finish the current question.

maheshgudi:
The can_skip_questions attribute is to be set True for students to be able to skip/cycle through questions.
If it is set as False, students wont be able to skip/cycle to the next question. closes #191

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -580,6 +580,10 @@ class Quiz(models.Model):
     view_answerpaper = models.BooleanField('Allow student to view their answer\
                                             paper', default=False)
 
+    can_skip_questions = models.BooleanField("Allow students to skip questions",
+                                             default=True
+                                             )
+
     objects = QuizManager()
 
     class Meta:
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -580,9 +580,8 @@ class Quiz(models.Model):
     view_answerpaper = models.BooleanField('Allow student to view their answer\
                                             paper', default=False)
 
-    can_skip_questions = models.BooleanField("Allow students to skip questions",
-                                             default=True
-                                             )
+    allow_skip = models.BooleanField("Allow students to skip questions",
+                                     default=True)
 
     objects = QuizManager()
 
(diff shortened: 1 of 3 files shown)


yaksh/models.py
```diff
@@ -580,6 +580,10 @@ class Quiz(models.Model):
     view_answerpaper = models.BooleanField('Allow student to view their answer\
                                             paper', default=False)
 
+    can_skip_questions = models.BooleanField("Allow students to skip questions",
```
prabhuramachandran: Can this be changed to `allow_skip`?

yaksh/templates/yaksh/question.html
```diff
@@ -199,14 +199,17 @@
          {% else %}
          <button class="btn btn-primary" type="submit" name="check" id="check" onClick="submitCode();">Check Answer <span class="glyphicon glyphicon-cog"></span></button>&nbsp;&nbsp;
          {% endif %}
+         {% if paper.question_paper.quiz.can_skip_questions %}
          {% if paper.unanswered.all|length != 1 %}
```
prabhuramachandran: This template value of `paper.unanswered.all` (I think it should be `questions_unanswered`) is probably wrong, please check this and bundle that check into the same, no need for two nested ifs here.

yaksh/models.py
maheshgudi: Resolved

yaksh/templates/yaksh/question.html
maheshgudi: Resolved

yaksh/models.py
```diff
@@ -580,6 +580,10 @@ class Quiz(models.Model):
     view_answerpaper = models.BooleanField('Allow student to view their answer\
                                             paper', default=False)
 
+    allow_skip = models.BooleanField("Allow students to skip questions",
+                                             default=True
```
prabhuramachandran: PEP8 formatting please.

maheshgudi: resolved

prabhuramachandran:
LGTM.

prabhuramachandran:
@ankitjavalkar -- please review this later when you return, for now I am merging this change.  If there are issues, please file a ticket so we can resolve them.

Issue #200, opened 2017-02-01
Add test for password reset


prabhuramachandran:
This has failed multiple times and no tests catch this.  It should be tested.

prabhuramachandran:
Fixed in #212.

PR #212, opened 2017-02-07
Add view test case for password reset and password change

ankitjavalkar:


prabhuramachandran:
Nice, thanks!  Please mark that it fixes the #200.

PR #214, opened 2017-02-08
removed max_length from stdio test case attributes

maheshgudi:
Removed max_length from stdio test case attributes because they restrict the user from adding more than 100 characters while providing with the expected input and output.


ankitjavalkar:
Merging on priority as this is affecting quiz creation

PR #221, opened 2017-02-09
Format the assertion and various error exceptions for Python Assertion evaluator

ankitjavalkar:


yaksh/python_assertion_evaluator.py
```diff
@@ -70,23 +71,41 @@ def check_code(self):
...
+                    stripped_tb
                     )
-        except TimeoutException:
-            raise
-        except Exception:
-            msg = traceback.format_exc(limit=0)
-            err = "Error in Test case: {0}".format(msg)
+
+
+            # err = "Error Traceback:\n{0}".format(stripped_tb)
```
prabhuramachandran: Please remove commented code.

yaksh/python_assertion_evaluator.py
```diff
@@ -70,23 +71,30 @@ def check_code(self):
...
             type, value, tb = sys.exc_info()
             info = traceback.extract_tb(tb)
             fname, lineno, func, text = info[-1]
             text = str(self.test_case)
+
+            # Get truncated traceback
+            err_tb_lines = traceback.format_exc().splitlines()
+            stripped_tb_lines = []
+            for line in err_tb_lines:
+                if '.py' not in line:
```
prabhuramachandran: seems dubious.

ankitjavalkar: OK. Would you prefer to use regex?

Reason: This is just to remove the file names / paths from the assertion since it displays internal files

prabhuramachandran: Not really, what if the exception had a .py inside it which was not a file.  Is there a test case that covers this specific condition?

prabhuramachandran: Instead, replace things like `File "/tmp/t.py", line 2, in foo` with a suitable change to the code if you really want to suppress the file name.  Do you have a specific case in mind?  You can use a regex for `File ".*\.py"` and replace it to what you want.  For example:
```
re.sub(r'File\s+".*?",\s+line', 'File filename, line', "File "/tmp/t.py", line 8, in <module>")
```

PR #224, opened 2017-02-13
Add smooth scroll to show errors in code questions

ankitjavalkar:


yaksh/templates/exam.html
```diff
@@ -26,6 +26,13 @@
     </nav><!-- /.navbar -->
 {% endblock %}
 {% block content %}
+  <script>
+    $(document).ready(function() {
+        var ele = document.getElementById('error_panel')
+        ele.scrollIntoView(true);
```
prabhuramachandran: This will give you JS errors since ele will be undefined, please write this better.

ankitjavalkar: fixed

Issue #213, opened 2017-02-07
Ordering of questions


prabhuramachandran:
It would be nice if the ordering of the questions was as chosen in the question paper instead of it being based on the key.

PR #228, opened 2017-02-22
Ordering Questions while creating Question Paper

adityacp:
This PR will add questions to Question Paper in the order of selection.
The Questions while displaying will not order by key and will show up according to selections made during creation of question paper.
This fixes issue #213

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -616,11 +616,11 @@ class Quiz(models.Model):
 ###############################################################################
 class QuestionPaperManager(models.Manager):
 
-    def _create_trial_from_questionpaper(self, original_quiz_id, fixed_que):
+    def _create_trial_from_questionpaper(self, original_quiz_id):
         """Creates a copy of the original questionpaper"""
         trial_questionpaper = self.get(quiz_id=original_quiz_id)
-        trial_questions = {"fixed_questions": fixed_que.get_fixed_questions(
-                            trial_questionpaper),
+        fixed_ques = trial_questionpaper.get_ordered_questions()
+        trial_questions = {"fixed_questions": fixed_ques,
                            "random_questions": trial_questionpaper
                            .random_questions.all()
                            }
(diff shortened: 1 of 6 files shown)


yaksh/models.py
```diff
@@ -748,14 +757,33 @@ def create_demo_quiz_ppr(self, demo_quiz, user):
...
     def __str__(self):
         return "Question Paper for " + self.quiz.description
 
 
 ###############################################################################
+class FixedQuestions(models.Model):
+    questionpaper = models.ForeignKey(QuestionPaper)
+    question = models.ForeignKey(Question)
+
+    def get_fixed_questions(self, qp):
```
maheshgudi: I think, it would be better to make this a model manager method.

yaksh/models.py
```diff
@@ -748,14 +757,33 @@ def create_demo_quiz_ppr(self, demo_quiz, user):
...
+        fixed_q = FixedQuestions.objects.filter(questionpaper=qp).order_by('pk')
+        questions = [que.question for que in fixed_q
+            if que.question.active == True]
+        return questions
+
+    def add_fixed_questions(self, qp, que):
```
maheshgudi: same here

yaksh/models.py
```diff
@@ -748,14 +757,33 @@ def create_demo_quiz_ppr(self, demo_quiz, user):
...
+        FixedQuestions.objects.create(questionpaper=qp, question=que)
+
+    def remove_fixed_questions(self, que_ids):
```
maheshgudi: same here

yaksh/models.py
```diff
@@ -619,8 +619,8 @@ class QuestionPaperManager(models.Manager):
     def _create_trial_from_questionpaper(self, original_quiz_id):
         """Creates a copy of the original questionpaper"""
         trial_questionpaper = self.get(quiz_id=original_quiz_id)
-        trial_questions = {"fixed_questions": trial_questionpaper
-                           .fixed_questions.all(),
+        fixed_ques = trial_questionpaper.get_ordered_questions()
```
prathamesh920: Fine for now, but can have more readable variable name.

PR #229, opened 2017-02-23
Allow to reattempt all types of questions

ankitjavalkar:


yaksh/models.py
```diff
@@ -988,7 +988,8 @@ def completed_question(self, question_id):
             questions and returns the next question.
         """
         next_question = self.next_question(question_id)
-        self.questions_answered.add(question_id)
+        if question_id not in self.questions_answered.all():
```
prabhuramachandran: The method should really be called `add_completed_question`.

ankitjavalkar: Done

yaksh/templates/yaksh/question.html
```diff
@@ -198,10 +210,15 @@
          {% elif question.type == "upload" %}
          <br><button class="btn btn-primary" type="submit" name="check" id="check" onClick="return validate();">Upload</button>&nbsp;&nbsp;
          {% else %}
-         <button class="btn btn-primary" type="submit" name="check" id="check" onClick="submitCode();">Check Answer <span class="glyphicon glyphicon-cog"></span></button>&nbsp;&nbsp;
+           {% if notification == None %}
```
prathamesh920: Check answer  button will not be displayed if there is a notification.
So in case I submit a blank code, a notification will be generated and 
check answer button will disappear.

ankitjavalkar: This has been fixed in the latest commit.

PR #231, opened 2017-02-28
Fix selenium tests:

ankitjavalkar:
Fix selenium tests:
    - Fix addition of answer to codemirror using selenium
    - Add custom exception when selenium fails
    - Add id to logout button on student interface
    - Fix imports for Python3 support in load_test.py

### DIFF
--- a/yaksh/live_server_tests/selenium_test.py
+++ b/yaksh/live_server_tests/selenium_test.py
@@ -8,6 +8,9 @@ from selenium.common.exceptions import WebDriverException
 import multiprocessing
 import argparse
 
+class SeleniumTestError(Exception):
+    pass
+
 class SeleniumTest():
     def __init__(self, url, quiz_name):
         self.driver = webdriver.Firefox()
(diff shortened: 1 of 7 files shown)


ankitjavalkar:
I still need to keep selenium in install_requires because of the following reason;

I get this error when selenium is not present in the install requires;
travis_without_selen1

This is not a problem with the module load_test, that file is present. As you can see below I have not installed selenium in this case (this is the only difference);
travis_without_selen2

After adding the selenium to the install requires, all the modules are found properly and everything passes successfully.
travis_with_selen1

ankitjavalkar:
The tests are not passing because the Travis is pulling from the master branch of FOSSEE/online_test. Here is a local branch that is passing https://travis-ci.org/ankitjavalkar/online_test/builds/206595857

yaksh/live_server_tests/selenium_test.py
```diff
@@ -24,9 +27,12 @@ def run_load_test(self, url, username, password):
...
             self.driver.close()
         except Exception as e:
-            with open("/tmp/yaksh_load_test_log.txt", "ab") as f:
-                f.write('Username: {0}\nError: {1}\n'.format(username, e))
             self.driver.close()
+            msg = ("An Error occurred while running the Selenium load"
+                " test on Yaksh!"
+                "Error:\n ".format(e))
+
+            raise SeleniumTestError()
```
prabhuramachandran: Need to call `raise SeleniumTestError(msg)` instead of not using the message.

ankitjavalkar: fixed

PR #244, opened 2017-03-10
Display error message while uploading questions

### LINKED ISSUE #232: No error message specifying correct file name when uploading questions
The upload functionality assumes that the questions are saved in `questions_dump.json`. 
I couldn't find any documentation specifying this and the upload page also did not show any errors when I tried to upload the JSON file with a different file name.

For better UI, it would be nice if the upload button is disabled or hidden until the user selects a file to upload or clicking it should directly open up the upload file dialog.

adityacp:
This PR fixes #232

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -318,18 +318,17 @@ class Question(models.Model):
             file_names = question.pop('files')
             test_cases = question.pop('testcase')
             que, result = Question.objects.get_or_create(**question)
-            if not result:
-                if file_names:
-                    que._add_files_to_db(file_names, file_path)
-                for test_case in test_cases:
-                    test_case_type = test_case.pop('test_case_type')
-                    model_class = get_model_class(test_case_type)
-                    new_test_case, obj_create_status = \
-                        model_class.objects.get_or_create(
-                            question=que, **test_case
-                        )
-                    new_test_case.type = test_case_type
-                    new_test_case.save()
+            if file_names:
+                que._add_files_to_db(file_names, file_path)
+            for test_case in test_cases:
+                test_case_type = test_case.pop('test_case_type')
+                model_class = get_model_class(test_case_type)
+                new_test_case, obj_create_status = \
+                    model_class.objects.get_or_create(
+                        question=que, **test_case
+                    )
+                new_test_case.type = test_case_type
+                new_test_case.save()
 
         if files_list:
             delete_files(files_list, file_path)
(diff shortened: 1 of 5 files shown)


yaksh/models.py
```diff
@@ -402,6 +402,9 @@ def read_json(self, file_path, user, files=None):
             with open(json_file, 'r') as q_file:
                 questions_list = q_file.read()
                 self.load_questions(questions_list, user, file_path, files)
+            return "Questions Uploaded Successfully"
+        else:
+            return "Please upload zip file with questions_dump.json in it."
```
maheshgudi: We are not raising an exception if the json data in the file is syntactically wrong. In my testing, uploading an improper json file will raise an ValueError exception with the information of where in the json file it found an error. Maybe we can display the error with its location.

yaksh/views.py
```diff
@@ -31,6 +30,7 @@
                 QuestionFilterForm, CourseForm, ProfileForm, UploadFileForm,\
                 get_object_form, FileForm, QuestionPaperForm
 from .settings import URL_ROOT
+from django.conf import settings
```
maheshgudi: Why can't we just import the FIXTURE_DIRS from online_test.settings instead of django.conf settings?

yaksh/models.py
maheshgudi: Hey incase of uploading a file with a bad file name, i.e. anything except filename questions_dump.json, will raise an exception. Hence this error message return is pretty useless.

yaksh/templates/yaksh/showquestions.html
```diff
@@ -10,18 +10,24 @@
...
+Questions</h4>
+<a href="{{URL_ROOT}}/exam/manage/download_demo_question">Sample Zip File</a>
+<br><br>
 <h4>Upload ZIP file for adding questions</h4>
 <form action="" method="post"  enctype="multipart/form-data">
 {% csrf_token %}
 {{ upload_form.as_p }}
-<button class="btn btn-primary" type="submit" name="upload" value="upload">Upload File <span class="glyphicon glyphicon-open"></span></button>
+<button class="btn btn-primary" type="submit" name="upload" value="upload">
+Upload File <span class="glyphicon glyphicon-open"></span></button>
```
adityacp: resolved added in docs.

yaksh/views.py
```diff
@@ -907,7 +908,8 @@ def show_all_questions(request):
                 if file_name[-1] == "zip":
```
adityacp: I will fix it later. This serves the purpose for now

PR #245, opened 2017-03-13
Migrations in script

### LINKED ISSUE #241: Improve yaksh script to auto-migrate

prathamesh920:
**Added migration commands to the script.**
Removed migrate  command with run-syncdb argument because this does not create
migration files, so further migrations become difficult.
Model changes are detected by make_migrations and the migrate command
applies those migrations. This is done every time when user runs a demo.
There is no harm in doing so.

**Added migrations folder to the app.**
This required because when we do migration for the first time, initial
migrations will be created only if migrations folder is present in the
installed app. Otherwise, running makemigrations will not detect the
initial migrations for that app.
Though we can exclusively specify the app name as the argument to the
makemigrations command.
But did this so that by default migrations will be applied.
**Note: Removed migrations from **.gitignore****

**Removed a function call to datetime now in default quiz start time.**
Every time when we run makemigrations, model changes are detected.
"Alter field start_date_time on quiz" was always detected, even if you
have not made any changes to the models!
This was happening because of default time value was datetime.now().
So every time when you ran migrations current datetime was set which will
always change.
Now removed parenthesis(), so that now function is not called.
And no unnecessary model changes will be detected

Closes #241


yaksh/scripts/cli.py
```diff
@@ -115,14 +111,18 @@ def create_demo(project_name='yaksh_demo', project_dir=CUR_DIR):
...
+        _migrate(project_name, command_path)
+        subprocess.call(loaddata_command, shell=True)
 
 
 def run_demo(project_name, top_dir):
     with _chdir(top_dir):
+        command_path = path.join(top_dir, 'manage.py')
+        _check_migrations(project_name, command_path)
+        _migrate(project_name, command_path)
         command = ("python manage.py runserver "
```
maheshgudi: We can add an additional command line argument for port number.

PR #249, opened 2017-03-14
Added mathjax support for quiz and answerpapers

### LINKED ISSUE #196: Add mathjax by default
Should make it easy to add formulae to questions.

maheshgudi:
Added Mathjax support for quiz and answerpapers.
closes #196

### DIFF
--- a/yaksh/templates/yaksh/user_data.html
+++ b/yaksh/templates/yaksh/user_data.html
@@ -5,7 +5,8 @@
 {% block content %}
 
 {% block script %}
-    <script src= "{{ URL_ROOT }}/static/yaksh/js/edit_question.js"></script>
+<script src= "{{ URL_ROOT }}/static/yaksh/js/edit_question.js"></script>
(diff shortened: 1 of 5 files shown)


ankitjavalkar:
Looks good, mergeable

yaksh/templates/yaksh/user_data.html
```diff
@@ -83,7 +84,7 @@
             {%endif%}
         </div>
     </div>
-    {% if question.type == "mcq" or question.type == "mcc" %}
+    {% if question.type != "code" %}
```
prathamesh920: I think checking question type explicitly is better, because the type can
be assignment upload or in future fill in the blanks. So, this may break.

maheshgudi: No, this doesn't break. Because in django, if it doesnt find a django variable while rendering, it will not raise an exception, it will just render with the variable value as none.

Issue #223, opened 2017-02-09
For STDIO report to the user what input we are giving to make it easier to debug


prabhuramachandran:

ankitjavalkar:
The output shows up as follows:
```
Incorrect answer
Given Input is
3 
Expected Output is u'6' 
Your output is u'3'
```
The code server runs in Python 3 hence the unicode representation is unavoidable. I'll check if that can be replaced/removed.

ankitjavalkar:
Fixed

PR #250, opened 2017-03-15
Multiple fixes in stdio_evaluator and models:

ankitjavalkar:
- Change the way stdio output is printed
- Fix minor errors when creating the uploaded file path in the models


ankitjavalkar:
Fixes #223

PR #253, opened 2017-03-20
Check in migration file to add field fixed_question_order to questionpaper

ankitjavalkar:


ankitjavalkar:
Merging on priority to fix broken master.

Issue #248, opened 2017-03-14
See and download assignments in answerpapers 


maheshgudi:
See and download assignments in view_answerpapers, grade user and monitor. Also moderator can bulk download all the students' uploaded assignments for a quiz question.

PR #268, opened 2017-03-31
Download assignments

adityacp:
This PR closes issue #248


yaksh/models.py
```diff
@@ -1289,7 +1290,26 @@ class AssignmentUpload(models.Model):
     user = models.ForeignKey(User)
     assignmentQuestion = models.ForeignKey(Question)
     assignmentFile = models.FileField(upload_to=get_assignment_dir)
+    question_paper = models.ForeignKey(QuestionPaper, blank=True, null=True)
 
+    def get_assignments(self, qp_id, que_id=None, user_id=None):
```
prathamesh920: Can we have a test for this method?

yaksh/views.py
```diff
@@ -494,20 +499,28 @@ def check(request, q_id, attempt_num=None, questionpaper_id=None):
             # validation
             if 'assignment' in request.FILES:
                 assignment_filename = request.FILES.getlist('assignment')
+                qp = QuestionPaper.objects.get(id=questionpaper_id)
             for fname in assignment_filename:
                 if AssignmentUpload.objects.filter(
                     assignmentQuestion=current_question,
-                    assignmentFile__icontains=fname, user=user).exists():
+                    assignmentFile__icontains=fname, user=user,
```
prathamesh920: If we save this query in a variable like assignment_files, then check assignment_files.exists()
Also, on assignment_files you can further query to get your "assign_file".
This will reduce lot of code and slight redundancy.

yaksh/test_models.py
```diff
@@ -1034,3 +1034,66 @@ def test_consolidate_answer_data(self):
...
+
+class AssignmentUploadTestCases(unittest.TestCase):
+    def setUp(self):
+        self.user = User.objects.get(username="demo_user")
+        self.user.first_name = "demo"
+        self.user.last_name = "user"
+        self.user.save()
+        course = Course.objects.create(name="Assignment course",
+                                   enrollment="Enroll Request", creator=self.user)
+        self.quiz = Quiz.objects.create(start_date_time=datetime(
```
prathamesh920: Can you use the course and quiz already created in module setup?

yaksh/test_models.py
```diff
@@ -1034,3 +1034,66 @@ def test_consolidate_answer_data(self):
...
+                                    self.questionpaper, self.question.id,
+                                    self.user.id
+                                    )
+        self.assertIn("upload.txt", assignment_files[0].assignmentFile.name)
+        self.assertEqual(assignment_files[0].question_paper, self.questionpaper)
+        actual_file_name = self.user.get_full_name().replace(" ", "_")
+        file_name = file_name.replace(" ", "_")
+        self.assertEqual(file_name, actual_file_name)
+
+    def test_get_assignments_for_quiz_files(self):
```
prathamesh920: Tests are almost same to ```test_get_assignments_for_user_files```
Can there be better implementation?

yaksh/views.py
```diff
@@ -910,14 +923,15 @@ def design_questionpaper(request, quiz_id, questionpaper_id=None):
...
             question_ids = request.POST.getlist('added-questions', None)
-            que_order = question_paper.fixed_question_order.split(",")
-            for qid in question_ids:
-                que_order.remove(qid)
-            if que_order:
-                question_paper.fixed_question_order = ",".join(que_order)
-            else:
-                question_paper.fixed_question_order = ""
-            question_paper.save()
+            if question_paper.fixed_question_order:
```
adityacp: It is required for question papers created without fixed question order. It throws error if tried to remove questions from question paper which does not have fixed question order.

yaksh/views.py
```diff
@@ -494,20 +499,29 @@ def check(request, q_id, attempt_num=None, questionpaper_id=None):
...
                     os.remove(assign_file.assignmentFile.path)
                     assign_file.delete()
                 AssignmentUpload.objects.create(user=user,
-                    assignmentQuestion=current_question, assignmentFile=fname
+                    assignmentQuestion=current_question, assignmentFile=fname,
+                    question_paper=qp
                     )
             user_answer = 'ASSIGNMENT UPLOADED'
             if not current_question.grade_assignment_upload:
+                new_answer = Answer(question=current_question, answer=user_answer,
```
adityacp: there should be a label that states whether assignment is correct or wrong.
- This works similar to code question, marks are given to the assignment question if evaluated with hook else it does not give marks.

all the attempts for a given upload questions should be shown like we do for code questions, currently only the last one is shown.
- It is not done because we replace files if it is already submitted.

also every time user answer shown is 'ASSIGNMENT UPLOADED', it should be the answer from the file.
- In this case I feel it is not possible everytime to read the user uploaded file, since uploaded file can be of any type instead it would be better to give download option.

so every time you are replacing the assignment file for a given question, that is fine, but at least in database there should be a record of what was submitted via file.
And same you can display on UI, this will make the UI very useful.
- If you are telling to read student uploaded file again I don't think that it will be possible for every case. So I felt giving download option is better.

yaksh/views.py
```diff
@@ -1403,3 +1427,33 @@ def download_course_csv(request, course_id):
...
+        folder_name = os.sep.join((name, os.path.basename(
+                        f_name.assignmentFile.name))
+                        )
+        zip_file.write(f_name.assignmentFile.path, folder_name
+                    )
+    zip_file.close()
+    zipfile_name.seek(0)
+    response = HttpResponse(content_type='application/zip')
+    response['Content-Disposition'] = '''attachment;\
+                                          filename={0}'''.format(
```
maheshgudi: Just add a .zip to the filename.

yaksh/views.py
prathamesh920: Uploaded Assignment are evaluated, so there is a chance of file read each time during evaluation.
So, why not simply read file content and store it in database, which will be very useful! Otherwise, every time I have to download to view!

prathamesh920: Yes marks are given to assignment questions, but for code question the UI looks better where there is green header panel for correct answer label. It will be better if the UI is consistent throughout.

yaksh/templates/404.html
```diff
@@ -1,5 +1,5 @@
 {% extends "base.html" %}
 
 {% block content %}
-The requested page does not exist.
+{{ exception }}
```
maheshgudi: <p>Could this exception come with a note like:</p>
It seems that you have encountered an error<br/>
Type of error - {{ exception }} <br/>
Please contact us at python@fossee.in

yaksh/templates/yaksh/grade_user.html
```diff
@@ -163,29 +163,41 @@
...
-                <div class="panel-heading">Correct answer:
+    {% if question.type == "upload" %}
+    {% if user_file_status %}
+        <a href="{{URL_ROOT}}/exam/manage/download/user_assignment/{{question.id}}/{{data.user.id}}/{{paper.question_paper.quiz.id}}">
+        <div class="panel">
+        Assignment File for {{ data.user.get_full_name.title }}
+        </div>
+        </a>
+    {% else %}
+        <h5>No Assignment submitted by {{ data.user.get_full_name.title }}</h5>
```
maheshgudi: I guess, the student should also be given a link to download his assignment file in view_answerpaper.html

yaksh/views.py
maheshgudi: @prathamesh920 
I can assure you that for at least two or three SDES quizzes, we had asked students to submit pdf files. Now opening/reading a pdf file and/or other similar file formats, and saving the content directly in the database will be an unnecessary feature addition to the code base right now.

Issue #270, opened 2017-04-03
Display Course, Quiz name in templates


adityacp:
Display course and quiz name in templates wherever required.
Change Design in Monitor.

Issue #266, opened 2017-03-30
Display Courses in Courses page properly


adityacp:
Show active courses, inactive courses and allotted courses separately.

adityacp:
Closing issue since the issue has been fixed

PR #271, opened 2017-04-04
Fix Grade User, Monitor Templates

adityacp:
This PR closes issues #270 #266

### DIFF
--- a/yaksh/views.py
+++ b/yaksh/views.py
@@ -284,9 +284,11 @@ def prof_manage(request, msg=None):
     user = request.user
     ci = RequestContext(request)
     if user.is_authenticated() and is_moderator(user):
-        question_papers = QuestionPaper.objects.filter(quiz__course__creator=user,
-                                                       quiz__is_trial=False
-                                                       )
+        question_papers = QuestionPaper.objects.filter(
+                                                Q(quiz__course__creator=user) |
+                                                Q(quiz__course__teachers=user),
+                                                quiz__is_trial=False
+                                                )
         trial_paper = AnswerPaper.objects.filter(user=user,
                                                  question_paper__quiz__is_trial=True
                                                  )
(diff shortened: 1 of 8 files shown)


yaksh/views.py
```diff
@@ -759,30 +761,31 @@ def show_statistics(request, questionpaper_id, attempt_number=None):
...
         return my_render_to_response('yaksh/monitor.html', context,
                                      context_instance=ci)
     # quiz_id is not None.
     try:
         q_paper = QuestionPaper.objects.filter(Q(quiz__course__creator=user) |
                                                Q(quiz__course__teachers=user),
                                                quiz__is_trial=False,
-                                               id=questionpaper_id).distinct()
+                                               quiz_id=quiz_id).distinct()
+        quiz = get_object_or_404(Quiz, id=quiz_id)
```
maheshgudi: Could this be in the beginning of the monitor view function? It seems to be oddly placed here.

yaksh/views.py
```diff
@@ -306,9 +308,9 @@ def prof_manage(request, msg=None):
         for paper in question_papers:
             answer_papers = AnswerPaper.objects.filter(question_paper=paper)
             users_passed = AnswerPaper.objects.filter(question_paper=paper,
-                    passed=True).count()
+                    passed=True).distinct().count()
```
prathamesh920: What ```distinct()``` will do here?
I think it can do nothing here, because our attempt number is always different.

yaksh/views.py
```diff
@@ -284,9 +284,11 @@ def prof_manage(request, msg=None):
...
     ci = RequestContext(request)
     if user.is_authenticated() and is_moderator(user):
-        question_papers = QuestionPaper.objects.filter(quiz__course__creator=user,
-                                                       quiz__is_trial=False
-                                                       )
+        question_papers = QuestionPaper.objects.filter(
+                                                Q(quiz__course__creator=user) |
+                                                Q(quiz__course__teachers=user),
+                                                quiz__is_trial=False
+                                                )
```
prathamesh920: Are you sure this will not return duplicate question papers?
Since you are using ```Q```, and a teacher can be in multiple courses.

adityacp:
Resolved comments

PR #267, opened 2017-03-31
Multiple changes to Course:

ankitjavalkar:
- Allow a start and end time for Course enrollment | Closes #242 
- Allow Course to be editable | Closes #261

PR #272, opened 2017-04-05
Course code feature

### LINKED ISSUE #219: Associate a general course code with a course
A course can set prerequisites using this course code.  Now instead of implementing this with strings, which is very fragile.  We should create a course template which is simply referred to in this particular course.  A moderator should be able to then edit and modify this course if needed by adding/removing quizzes, editing quizzes etc.

ankitjavalkar:
This request is dependent on #267 Kindly merge only after the former is merged.
Closes #219

### DIFF
--- a/yaksh/forms.py
+++ b/yaksh/forms.py
@@ -274,6 +274,17 @@ class QuestionFilterForm(forms.Form):
 class CourseForm(forms.ModelForm):
     """ course form for moderators """
 
+    def save(self, commit=True, *args, **kwargs):
+        instance = super(CourseForm, self).save(commit=False)
+        if instance.code:
+            instance.hidden = True
+        else:
+            instance.hidden = False
+
+        if commit:
+            instance.save()
+        return instance
+
     class Meta:
         model = Course
         exclude = ['creator', 'requests', 'students', 'rejected',
--- a/yaksh/forms.py
+++ b/yaksh/forms.py
@@ -288,7 +288,7 @@ class CourseForm(forms.ModelForm):
     class Meta:
         model = Course
         exclude = ['creator', 'requests', 'students', 'rejected',
-            'created_on', 'is_trial', 'teachers']
+            'created_on', 'is_trial', 'hidden', 'teachers']
 
 
 class ProfileForm(forms.ModelForm):
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -112,6 +112,8 @@ class CourseManager(models.Manager):
         trial_course.enroll(False, user)
         return trial_course
 
+    def get_hidden_courses(self, code):
+        return self.filter(code=code, hidden=True)
 
 ###############################################################################
 class Course(models.Model):
@@ -119,6 +121,8 @@ class Course(models.Model):
     name = models.CharField(max_length=128)
     enrollment = models.CharField(max_length=32, choices=enrollment_methods)
     active = models.BooleanField(default=True)
+    code = models.CharField(max_length=128, null=True, blank=True)
+    hidden = models.BooleanField(default=False)
     creator = models.ForeignKey(User, related_name='creator')
     students = models.ManyToManyField(User, related_name='students')
     requests = models.ManyToManyField(User, related_name='requests')
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -49,7 +49,7 @@ question_types = (
 
 enrollment_methods = (
     ("default", "Enroll Request"),
-    ("open", "Open Course"),
+    ("open", "Open Enrollment"),
     )
 
 test_case_types = (
--- a/yaksh/templates/yaksh/quizzes_user.html
+++ b/yaksh/templates/yaksh/quizzes_user.html
@@ -29,7 +29,7 @@ No Courses to display
                 <h4><b><u> {{ course.name }} by {{ course.creator.get_full_name }}</u></b></h4>
             </div>
              <div class="col-md-4">
-                {% if course.hidden %}<span class="label label-info">Hidden</span>
+                {% if course.hidden %}<span class="label label-info">Open Course</span>
                 {% endif %}
                 {% if user in course.requests.all %} <span class="label label-warning">Request Pending </span>
                 {% elif user in course.rejected.all %}<span class="label label-danger">Request Rejected</span>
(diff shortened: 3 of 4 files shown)


yaksh/views.py
```diff
@@ -117,14 +117,25 @@ def user_logout(request):
 def quizlist_user(request, enrolled=None):
     """Show All Quizzes that is available to logged-in user."""
     user = request.user
-    if enrolled is not None:
+    ci = RequestContext(request)
+
+    if request.method == "POST":
+        course_code = request.POST.get('course_code')
+        hidden_courses = Course.objects.get_hidden_courses(code=course_code)
+        courses = hidden_courses if hidden_courses else None
```
maheshgudi: You could use django messages

yaksh/models.py
```diff
@@ -111,13 +111,17 @@ def create_trial_course(self, user):
...
+        return self.filter(code=code, hidden=True)
 
 ###############################################################################
 class Course(models.Model):
     """ Course for students"""
     name = models.CharField(max_length=128)
     enrollment = models.CharField(max_length=32, choices=enrollment_methods)
     active = models.BooleanField(default=True)
+    code = models.CharField(max_length=128, null=True, blank=True)
+    hidden = models.BooleanField(default=False)
```
maheshgudi: If a moderator checks hidden, and doesn't add any course code for it, the course is basically inaccessible to everyone.

yaksh/forms.py
```diff
@@ -274,6 +274,17 @@ def __init__(self, *args, **kwargs):
 class CourseForm(forms.ModelForm):
     """ course form for moderators """
 
+    def save(self, commit=True, *args, **kwargs):
```
prathamesh920: I think having two fields i.e. **code** and **hidden** for this feature is slightly increasing the complexity.
I mean hidden field is of no use for the moderator. **hidden** field is not changeable by the moderator, it simply depends on the course code. So each time we need to handle hidden field as per the course code. 

We can simply have a **code** field, that solely determines everything for this feature.
If you want to display only those courses that do not have code then the following can be done:
```Course.objects.exclude(code__isnull=True)```
I think this will reduce lot of your lines of code and makes it simpler.

File: yaksh/templates/yaksh/quizzes_user.html [hunk 4]
```diff
@@ -1,6 +1,21 @@
...
+<div class="row well">
+    <form action="{{ URL_ROOT }}/exam/quizzes/" method="post" id="custom-search-form" class="form-search form-horizontal pull-right">
+        {% csrf_token %}
+        <div class="col-md-12">
+            <div class="input-group">
+                <span class="input-group-addon" id="basic-addon1">Search Course</span>
+                <input type="text" name="course_code" class="form-control" placeholder="Course Code">
+                <span class="input-group-btn">
+                    <button class="btn btn-default" type="submit">Search</button>
+                    <button class="btn btn-default" type="button" name="button" onClick='location.replace("{{URL_ROOT}}/exam/quizzes/");'>Cancel</button>
```
prathamesh920: **Cancel button** is reloading the home page.
Can remove the button!
Else replace cancel with clear and only clear the search box without reloading.

yaksh/views.py
prathamesh920: Yes I agree, will be better.

yaksh/models.py
prathamesh920: Its been handled in forms where save is overridden, **hidden** is always saved as false if there is no course code, irrespective of user selection.

yaksh/templates/yaksh/quizzes_user.html
ankitjavalkar: The cancel button refreshes the page because after the search is executed only hidden courses are displayed. Hitting cancel will reload and display the regular, open, non-code courses

yaksh/models.py
ankitjavalkar: @maheshgudi @prathamesh920 The hidden field itself is supposed to be hidden, will make the necessary changes

yaksh/forms.py
ankitjavalkar: I personally feel that depending on s Character field for decision making is poor design in this case. These 2 fields allow greater flexibility because in case a design decision is made where Courses with codes are not supposed to be hidden from plain sight, then refactoring becomes a pain.
Moreover, this field will now allow us to hide special courses that students should not have access to like the Demo Yaksh Course.

yaksh/views.py
ankitjavalkar: Agreed will make the change

ankitjavalkar:
#219 Just specifies that prerequisites could be specified using course codes.

yaksh/views.py
maheshgudi: Also, could there be a clear way for a student to distinguish between Open courses and hidden courses. With the current implementation, The open courses underneath the search bar would be pretty confusing for the students. Just an heading saying open courses could be enough.

yaksh/views.py
```diff
@@ -117,14 +117,25 @@ def user_logout(request):
...
+        title = 'Search'
+
+    elif enrolled is not None:
```
maheshgudi: Why do we have the search bar on the enrolled courses page? Having the search bar on two different pages is a bit unnecessary.

yaksh/forms.py
maheshgudi: I have to side with @ankitjavalkar  on this. We might need this feature in the future.

yaksh/views.py
maheshgudi: @ankitjavalkar: Could this be implemented?

ankitjavalkar: Done. Though I think that a common template for such purposes leads to dirty If-Else statements like the ones that I had to add.

yaksh/views.py
ankitjavalkar: Done.

yaksh/migrations/0006_auto_20170424_1015.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.5 on 2017-04-24 10:15
+from __future__ import unicode_literals
+
+from django.db import migrations, models
```
ankitjavalkar: We need to finalise a naming convention ASAP and update all PRs accordingly, please provide suggestions here https://github.com/FOSSEE/online_test/issues/278

I will rename as per the final decision made in the above issue.

ankitjavalkar:
Removed migrations from this PR

PR #273, opened 2017-04-07
Email verification during registration

### LINKED ISSUE #264: Add a email verification workflow for new user registration

adityacp:
This PR closes issue #264.


yaksh/views.py
```diff
@@ -88,15 +91,20 @@ def user_register(request):
...
+    context = {}
     if request.method == "POST":
         form = UserRegisterForm(request.POST)
         if form.is_valid():
             data = form.cleaned_data
-            u_name, pwd = form.save()
-            new_user = authenticate(username=u_name, password=pwd)
-            login(request, new_user)
-            return my_redirect("/exam/quizzes/")
+            if settings.IS_DEVELOPMENT:
```
ankitjavalkar: Try to figure out if this is possible through a middleware.

yaksh/views.py
```diff
@@ -1403,3 +1421,45 @@ def download_course_csv(request, course_id):
     for student in students:
         writer.writerow(student)
     return response
+
+def activate_user(request, key):
+    ci = RequestContext(request)
+    profile = get_object_or_404(Profile, activation_key=key)
```
maheshgudi: So if I click on the link that I received in the mail for the second time(after verifying the first time), I am again redirected to the **'enter-email-to-verify'** page. Why is it implemented that way? Since the activation key is unique to a user and the activation key constitutes a part of the url meant for activation, I should just be redirected to the **'your-email-is-already-verified'** page.

yaksh/models.py
```diff
@@ -251,6 +251,9 @@ class Profile(models.Model):
         default=pytz.utc.zone,
         choices=[(tz, tz) for tz in pytz.common_timezones]
     )
+    is_email_verified = models.BooleanField(default=False)
+    activation_key = models.CharField(max_length=40, blank=True, null=True)
```
maheshgudi: Is it necessary to hold onto the activation key in the database once the verification is done?

yaksh/models.py
```diff
@@ -251,6 +251,9 @@ class Profile(models.Model):
...
+    key_expiry_time = models.DateTimeField(blank=True, null=True)
```
maheshgudi: Same here.

yaksh/middleware/user_time_zone.py
```diff
@@ -5,10 +5,10 @@
 
 class TimezoneMiddleware(object):
     """ Middleware to get user's timezone and activate timezone 
-        if user timezone is not available default value 'UTC' is activated """
+        if user timezone is not available default value 'Asia/Kolkata' is activated """
     def process_request(self, request):
         user = request.user
-        user_tz = 'UTC'
+        user_tz = 'Asia/Kolkata'
```
maheshgudi: Nothing wrong here, but is it really necessary for the default date time value to be Asia/Kolkata? UTC is a time standard, not a time zone. Every timezone is derived wrt UTC(IST being UTC+05:30). It makes more sense for the default value to be UTC.

yaksh/models.py
adityacp: It is not necessary to keep the key but storing the key will not cause any issue.

yaksh/models.py
adityacp: Same as above

yaksh/middleware/user_time_zone.py
adityacp: Setting timezone to Asia/Kolkata is necessary because it makes easier to set the time for the quizzes. If it kept to UTC then it becomes bit difficult to set quiz timings

adityacp:
Tests are failing because migrations are not added in this PR. This issue will be fixed once PR #282 is merged.
All the comments have resolved

online_test/settings.py
```diff
@@ -102,6 +102,7 @@
 
 MEDIA_ROOT = os.path.join(BASE_DIR, "yaksh", "data")
 
+# Set this varable to <True> if smtp-server is not allowing to send email.
```
ankitjavalkar: Set this variable to <True> if smtp-server is not sending mails

PR #277, opened 2017-04-24
Course clone

ankitjavalkar:


prabhuramachandran:
You should test this better, as per issue #287.  Please do this later.

Issue #281, opened 2017-04-26
Tests fail if migrations are not included in the pull request


ankitjavalkar:
Raised based on @adityacp comment in #278

PR #282, opened 2017-04-26
Dont use migrations while testing

ankitjavalkar:
* Added a test_settings file to avoid the production migrations
* Updated travis.yml


ankitjavalkar:
Closes #281

PR #288, opened 2017-04-28
Add a negative test case to check attributes are duplicated

### LINKED ISSUE #287: Write negative test that fails if new attributes are not duplicated

ankitjavalkar:
closes #287

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -151,21 +151,21 @@ class Course(models.Model):
 
     objects = CourseManager()
 
+    def _create_duplicate_instance(self, creator, course_name=None):
+        new_course = self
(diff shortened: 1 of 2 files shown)


yaksh/models.py
```diff
@@ -151,21 +151,21 @@ class Course(models.Model):
 
     objects = CourseManager()
 
+    def _create_duplicate_instance(self, creator, course_name=None):
+        new_course = self
+        new_course.id = None
+        new_course.name = course_name if course_name else self.name
```
prathamesh920: new_course.name is already self.name because 1st line in the method is ```new_course = self```
Actually no harm.

PR #290, opened 2017-05-03
Email verification testing

adityacp:
Added views tests for email verification.


yaksh/test_views.py
```diff
@@ -97,6 +98,31 @@ def test_view_profile_get_for_user_with_profile(self):
...
+        self.client.login(
+            username=self.user2.username,
+            password=self.user2_plaintext_pass
+        )
+        post_response = self.client.post(reverse('yaksh:new_activation'),
+                data={'email':self.user2.email}
+            )
+        subject = mail.outbox[0].subject.replace(" ", "_")
+        activation_key = mail.outbox[0].body.split("\n")[2].split("/")[-1]
+        get_response = self.client.get(reverse('yaksh:activate',
```
maheshgudi: Although this is fine, my worry is on a machine with a SMTP server, this test will actually send a mail to the email id of the test user. But if we use a different email backend provided by Django, (like dummy or file backends) for development, it would be better. Maybe have a look into this?

yaksh/send_emails.py
```diff
@@ -40,20 +43,7 @@ def send_user_mail(user_mail, key):
...
-        smtpserver.ehlo()
-        smtpserver.esmtp_features['auth']='LOGIN DIGEST-MD5 PLAIN'
-
-        smtpserver.login(user, pwd)
-        header = 'To:{0}\nFrom:{1}\nSubject:{2}\n'.format(to,
-                    settings.SENDER_EMAIL, subject)
-        message = '{0}\n{1}\n\n'.format(header, message)
-        smtpserver.sendmail(user, to, message)
-        smtpserver.close()
+        send_mail(subject, message, settings.SENDER_EMAIL, [to])
```
maheshgudi: You can set connection parameter as dummy backend for development

yaksh/test_views.py
adityacp: Resolved

yaksh/send_emails.py
adityacp: Resolved

PR #296, opened 2017-05-16
Fix email verfication bug

adityacp:
Small bug fix for email verification


yaksh/forms.py
```diff
@@ -150,7 +150,8 @@ def save(self):
             new_profile.key_expiry_time = timezone.now() + \
                                         timezone.timedelta(minutes=20)
         new_profile.save()
-        return u_name, pwd, new_user.email, new_profile.activation_key
+        user_activation_key = Profile.objects.get(user=new_user).activation_key
```
maheshgudi: Isn't <code> user_activation_key </code> the same as <code>new_profile.activation_key</code>.

maheshgudi: Also what happens to <code>user_activation_key</code> if <code>settings.IS_DEVELOPMENT</code> is True? Will it send a None value or will that raise an error?

yaksh/models.py
```diff
@@ -293,7 +293,7 @@ class Profile(models.Model):
         choices=[(tz, tz) for tz in pytz.common_timezones]
     )
     is_email_verified = models.BooleanField(default=False)
-    activation_key = models.CharField(max_length=40, blank=True, null=True)
+    activation_key = models.CharField(max_length=255, blank=True, null=True)
```
maheshgudi: this should also have <code>default=True</code> as an argument.

yaksh/views.py
```diff
@@ -1543,7 +1543,10 @@ def new_activation(request, email=None):
         user.profile.key_expiry_time = timezone.now() + \
                                 timezone.timedelta(minutes=20)
         user.profile.save()
-        success, msg = send_user_mail(user.email, user.profile.activation_key)
+        new_user_data = User.objects.get(email=email)
+        success, msg = send_user_mail(new_user_data.email,
+                                      new_user_data.profile.activation_key
```
maheshgudi: Again, <code>new_user_data.profile.activation_key</code> is the same as <code>user.profile.activation_key</code>

yaksh/forms.py
adityacp: No, user_activation_key is not same as new_profile.activation_key. Suppose the key is generated with length more than 40 (which was length of activation_key field earlier) it used to save only 40 chars and activation key in email will contain full key so while verifying key it throws error saying key not matched. In order to avoid this situation this is done.

adityacp: If `settings.IS_DEVELOPMENT` is set True then `user_activation_key` will be blank entry. which is checked in views to avoid sending mail if `user_activation_key` is blank

yaksh/models.py
adityacp: is_email_verified field cannot have default=True since in production it will not send email, because in views the condition is there which will check this variable and it will find this variable to be True and will not send mail

yaksh/views.py
adityacp: Same reason as mentioned in earlier comment.

ankitjavalkar:
Looks good. Can merge.

PR #302, opened 2017-06-09
Search questions using tag

### LINKED ISSUE #301: Using question tags to search questions
Question tags can be used to filter/search questions.

maheshgudi:
- Questions can be searched using tags.
- This can be added for searching questions during  creating question papers in the future.
- closes #301

screenshot from 2017-06-09 16-32-25


yaksh/views.py
```diff
@@ -1071,9 +1072,24 @@ def show_all_questions(request):
             else:
                 context["msg"] = "Please select atleast one question to test"
 
+        if request.POST.get('question_tags'):
+            question_tags = request.POST.getlist("question_tags")
+            search_tags = []
+            for tags in question_tags:
+                search_tags.extend(re.split('[; |, |\*|\n]',tags))
```
ankitjavalkar: Can this be pushed to client-side JS and have the JS send a JSON string? We can then convert the JSON to a dict/list in a easier fashion.

yaksh/views.py
```diff
@@ -1071,9 +1072,24 @@ def show_all_questions(request):
...
+            search_result = Question.objects.filter(tags__name__in=search_tags,
+                                                    user=user).distinct()
+            context['search_result'] = search_result
+
     questions = Question.objects.filter(user_id=user.id, active=True)
     form = QuestionFilterForm(user=user)
+    user_tags = Question.objects.filter(user=user)\
```
ankitjavalkar: I personally feel that the '\' line break should be avoided.

Please store the ```Question.objects.filter(user=user)``` in a variable and then filter further so that you are within the 80 cols

yaksh/views.py
```diff
@@ -1071,9 +1072,24 @@ def show_all_questions(request):
...
+                 .values_list('tags', flat=True).distinct()
+    
+
```
ankitjavalkar: Everyone needs a little space for themselves. However code does not, why the two lines?

yaksh/views.py
maheshgudi: I dont think it makes a whole lot of difference. Anyhow I have to split the string, even if the string is returned from js.

maheshgudi:
Resolved

ankitjavalkar:
Looks good

yaksh/templates/yaksh/showquestions.html
```diff
@@ -38,18 +38,55 @@ <h5 style="padding-left: 20px;">Filters</h5>
...
+{% if not search_result %}
 {% if questions %}
 <h5><input id="checkall" type="checkbox"> Select All </h5>
 {% for i in questions %}
 <input type="checkbox" name="question" value="{{ i.id }}">&nbsp;&nbsp;<a href="{{URL_ROOT}}/exam/manage/addquestion/{{ i.id }}">{{ i }}</a><br>
 {% endfor %}
 {% endif %}
+{% else %}
+<h5><input id="checkall" type="checkbox"> Select All </h5>
+{% for i in search_result %}
```
prathamesh920: can we have question in search_result instead of i

yaksh/templates/yaksh/showquestions.html
```diff
@@ -38,18 +38,55 @@ <h5 style="padding-left: 20px;">Filters</h5>
...
+    </select>
+    </div>
+    </div>
+     </div>
+<br><br>
+<button class="btn btn-primary" type="button" onClick='location.replace("{{URL_ROOT}}");'>
+ Clear Filters</button>
 
 <div id="filtered-questions">
+{% if not search_result %}
```
prathamesh920: if no such tags then user will not know! It will display all questions, giving an impression that all questions belong to the searched tags.

maheshgudi:
Resolved all isssues

PR #303, opened 2017-06-21
Adds tabbed view for error output

maheshgudi:
- Adds tabbed view of error output for testcases during exam, instead of a long list of error output for each testcase.
- Fixes word wrap issue with stdIO error output
peek 2017-06-22 01-51


maheshgudi:
Closing due to unpopularity among the team.

PR #304, opened 2017-07-14
Adds search feature for Profile model in admin interface.

maheshgudi:
In light of recent events, it is advisable to have a search feature for the model Profile in the admin interface, so that the admin can manually search and verify specific email ids, in case of some unforeseeable failure. 
Adding search feature for that.


ankitjavalkar:
Merging as this is a minor feature

PR #305, opened 2017-07-21
Add a has_profile decorator

ankitjavalkar:


yaksh/decorators.py
```diff
@@ -1,20 +1,47 @@
...
+    This decorator is used to check if the user account has a profile.
+    If the user does not have a profile then redirect the user to
+    profile edit page.
+    """
+
+    def _wrapped_view(request, *args, **kwargs):
+        if user_has_profile(request.user):
+            return func(request, *args, **kwargs)
+        ci = RequestContext(request)
+        template = 'manage.html' if 'moderator' in request.user.groups.all() else 'user.html'
```
maheshgudi: Why <code>request.user.groups.all()</code>, could it just check if user is in the moderator group?

ankitjavalkar: This is doing exactly that.
I check if a user is associated with a group called 'moderator'. He may or may not be associated with other groups.

maheshgudi: Ok. Resolved.

prathamesh920: Can we use the **is_moderator** function here, that is already been defined in views?

yaksh/decorators.py
```diff
@@ -1,20 +1,47 @@
...
+        form = ProfileForm(user=request.user, instance=None)
```
prathamesh920: This should work without the instance argument.

yaksh/test_views.py
```diff
@@ -206,6 +207,30 @@ def test_edit_profile_get(self):
...
+        If profile exists a editprofile.html template will be rendered
+        """
+        self.client.login(
+            username=self.user2.username,
+            password=self.user2_plaintext_pass
+        )
+        response = self.client.get(reverse('yaksh:edit_profile'))
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, 'yaksh/editprofile.html')
+
```
prathamesh920: We can better tests like when user logins with or without profile.

yaksh/decorators.py
ankitjavalkar: This leads to cyclic import causing an ImportError since views also imports functions from decorators and decorator will import from views.

Adding the underlying condition from ```is_moderator``` here until a better solution is found

yaksh/decorators.py
ankitjavalkar: Done

yaksh/test_views.py
ankitjavalkar: Done, added tests for all 3 functions where the has_profile decorator is being used

prathamesh920:
LGTM

PR #309, opened 2017-07-27
Fix bug to prevent students from attempting pre-enrolled deactivated course

### LINKED ISSUE #308: Deactivated Quizzes show up in 'Enrolled Courses' page
- Enrolled Courses should explicitly show a Course (which the student has already enrolled for) but keep it in disabled state
- Home page (where un-enrolled courses show up) will not show the deactivated courses at all.
- The Quizzes, Answerpapers, etc should not be visible at all.

ankitjavalkar:
Fixes #308

### DIFF
--- a/yaksh/views.py
+++ b/yaksh/views.py
@@ -375,7 +375,7 @@ def start(request, questionpaper_id=None, attempt_num=None):
     if not quest_paper.quiz.course.is_enrolled(user):
         raise Http404('You are not allowed to view this page!')
     # prerequisite check and passing criteria
-    if quest_paper.quiz.is_expired():
+    if quest_paper.quiz.is_expired() and not quest_paper.quiz.course.active:
         if is_moderator(user):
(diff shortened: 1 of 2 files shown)


yaksh/views.py
```diff
@@ -375,7 +375,7 @@ def start(request, questionpaper_id=None, attempt_num=None):
     if not quest_paper.quiz.course.is_enrolled(user):
         raise Http404('You are not allowed to view this page!')
     # prerequisite check and passing criteria
-    if quest_paper.quiz.is_expired():
+    if quest_paper.quiz.is_expired() and not quest_paper.quiz.course.active:
```
prathamesh920: This should be OR condition.
```
if quiz is expired or course inactive:
    can't attempt
```

prathamesh920: Also, we can have tests for the same.

ankitjavalkar: Done

PR #311, opened 2017-08-02
Display all hidden courses with pending enrollment requests

ankitjavalkar:
Note: PEP8 issues have been fixed in earlier PR #309


ankitjavalkar:
Noted. Fixed the query to avoid such issues. I went with improving the query instead of drastic changes to the UI

prathamesh920:
LGTM

PR #324, opened 2017-08-30
Minor Bug fix (Anonymous User and Multiple Answerpaper errors) 

### LINKED ISSUE #322: Multiple Answerpaper objects returned bug
This bug takes place when a student has two answer paper objects containing the same attempt number.

### LINKED ISSUE #323: AnonymousUser is not iterable bug
This bug mostly happens when a user tries to quit an exam (manually entering the quit url )without authentication.

maheshgudi:
Fixes anonymous user not iterable bug and Multiple answerpaper objects returned bugs.
Closes #323 
Closes #322

### DIFF
--- a/yaksh/views.py
+++ b/yaksh/views.py
@@ -393,8 +393,8 @@ def start(request, questionpaper_id=None, attempt_num=None):
         if is_moderator(user):
             return redirect("/exam/manage")
         return redirect("/exam/quizzes")
+    attempt_number = 1 if not last_attempt else last_attempt.attempt_number + 1
     if attempt_num is None:
-        attempt_number = 1 if not last_attempt else last_attempt.attempt_number +1
         context = {'user': user, 'questionpaper': quest_paper,
                    'attempt_num': attempt_number}
         if is_moderator(user):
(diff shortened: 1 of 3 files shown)


yaksh/views.py
```diff
@@ -402,12 +402,23 @@ def start(request, questionpaper_id=None, attempt_num=None):
...
+        attempted_paper = AnswerPaper.objects.filter(user=user,
+                                                     question_paper=quest_paper,
+                                                     attempt_number=attempt_num
+                                                     )
+        if not attempted_paper:
+            ip = request.META['REMOTE_ADDR']
+            if not hasattr(user, 'profile'):
+                msg = 'You do not have a profile and cannot take the quiz!'
+                raise Http404(msg)
+            new_paper = quest_paper.make_answerpaper(user, ip, attempt_num)
```
prathamesh920: Nice catch,
I think here we need to take care of one more thing, if in case attempt number is any random number then too answer paper will be created. Because we are only checking if the attempt number is in database table, if not then create paper.
The attempt number should always be plus one then the last attempt number.

yaksh/views.py
```diff
@@ -402,12 +402,28 @@ def start(request, questionpaper_id=None, attempt_num=None):
         return my_render_to_response('yaksh/intro.html', context,
                                      context_instance=ci)
     else:
-        ip = request.META['REMOTE_ADDR']
-        if not hasattr(user, 'profile'):
-            msg = 'You do not have a profile and cannot take the quiz!'
+        attempted_paper = AnswerPaper.objects.filter(user=user,
```
prathamesh920: A suggestion, you can have a get over here instead of filter because in future there will never be multiple answer papers with same attempt number for a given question paper.
Using try except will help, if get fails with DoesNotExist exception then you can create an answer paper for that attempt number. If MultipleObjectsReturned exception then you can take the last answer paper for that attempt.

maheshgudi: resolved

yaksh/views.py
```diff
@@ -420,11 +420,11 @@ def start(request, questionpaper_id=None, attempt_num=None):
         if is_moderator(user):
             return redirect("/exam/manage")
         return redirect("/exam/quizzes")
+    if not last_attempt:
+        attempt_number = 1
+    else:
+        last_attempt.attempt_number + 1
```
prabhuramachandran: This statement looks to be meaningless as it does not explicitly set anything, did you mean:
`attempt_number = last_attempt.attempt_number + 1`?  Why wasn't this caught by the test?

maheshgudi: Resolved.

PR #325, opened 2017-08-31
Installs Firefox v46.0 on Travis CI

maheshgudi:
Selenium v2.53.6 seems to work perfectly with Firefox v46.0. Travis CI is expected to keep upgrading their default linux distribution, which might not have Firefox v46.0 in it's repository. Hence we enforce Travis CI to install Firefox v46.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,3 +25,6 @@ script:
 after_success:
   - coverage combine
   - coverage report
+
+addons:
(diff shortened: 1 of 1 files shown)


prathamesh920:
LGTM
In future we need to update selenium and travis configuration related to OS. We may discuss this and add a ticket.

Issue #285, opened 2017-04-28
Chat interface to allow moderator-to-student communication


ankitjavalkar:
Provide a chat interface/side-bar that allows the following;

* Simple one-to-one, duplex chat between moderator and students.
* Broadcast chat from moderator to all students.
* Prevent student from chatting amongst themselves

PR #327, opened 2017-09-06
Yaksh chat

adityacp:
Chat system for Yaksh.
Supports Broadcast chatting between moderator and students.
Closes issue #285


yaksh/consumers.py
```diff
@@ -0,0 +1,90 @@
...
+
+    if set(data.keys()) != set(('sender_id', 'message')):
+        message.reply_channel.send({'close': True})
+        return
+
+    if data:
+        context = {}
+        chat_dict = {}
+        message = room.messages.create(**data)
+        sender_tz = User.objects.get(id=data['sender_id']).profile.timezone
```
maheshgudi: The timezone while saving should be saved as GMT. And while rendering the time, the time should be rendered according to the user's timezone. This doesn't happen currently

yaksh/static/yaksh/js/chat.js
```diff
@@ -0,0 +1,166 @@
...
+    var chatsock;
+    var label;
+    var timer;
+    var page_title = document.title;
+    var course_id = $("#course_id").val();
+
+    $("#main").on("click", function(){
+        // Open chat window
+        var room_url = window.location.protocol + "//" +
+                    window.location.host + "/exam/new/chat/" + course_id;
```
maheshgudi: I am not sure if the <code>room_url</code> should be a part of the client side. Or at least it should encrypted in some form. Exposing the url to the client can increase vulnerability.

yaksh/static/yaksh/js/chat.js
```diff
@@ -0,0 +1,166 @@
...
+        ajax_request(room_url);
+    });
+
+    function ajax_request(chat_url) {
+        // Get messages from ajax request
+        $.ajax({
+            url: chat_url,
+            timeout:30000,
```
maheshgudi: 30000 seconds?

yaksh/static/yaksh/js/chat.js
```diff
@@ -0,0 +1,166 @@
...
+                message: $('#message').html(),
+            }
+            chatsock.send(JSON.stringify(message));
+        }
+        $("#message").empty();
+        return false;
+    });
+
+    // Create divs for showing sent and received messages
+    var div_sent = "<div class='row msg_container base_sent'>\n"+
```
maheshgudi: Cant this be done with <code>document.createElement</code>? This is just horrible!

yaksh/static/yaksh/js/chat.js
adityacp: I feel it is okay. Just need to make sure from views that no intruder can get messages. I feel I covered checks for user to get messages in views function

ankitjavalkar:
Need to add feature to changelog

prathamesh920:
Closing for now.
Will reopen with 0.8 release.

PR #331, opened 2017-09-20
Fix change password and other bug fixes.

maheshgudi:
1. Fixes change password error. Change password link on Moderator's dashboard was broken
2. Recurring 'NoneType has no attribute encode' error fixed.


ankitjavalkar:
LGTM

PR #336, opened 2017-09-28
New js editor

adityacp:
This PR adds new js editor TinyMCE.
Earlier js editor requires a license. So I added js libraries for the new editor.


ankitjavalkar:
Need to add the feature to the CHANGELOG list

PR #341, opened 2017-10-04
Hotfix7

adityacp:
This PR includes fix for the following:-
1. Update answer paper marks as per latest attempts for non-coding questions.
2. Show last submitted answer during question revisits.
3. Prevent storing answers during skipping of questions.

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -1285,7 +1285,7 @@ class AnswerPaper(models.Model):
         return q_a
 
     def get_latest_answer(self, question_id):
-        return self.answers.filter(question=question_id).order_by("-id").last()
+        return self.answers.filter(question=question_id).order_by("id").last()
 
     def get_questions(self):
         return self.questions.filter(active=True)
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -1308,8 +1308,7 @@ class AnswerPaper(models.Model):
             return self.time_left() > 0
 
     def get_previous_answers(self, question):
-        if question.type == 'code':
-            return self.answers.filter(question=question).order_by('-id')
+        return self.answers.filter(question=question).order_by('-id')
 
     def validate_answer(self, user_answer, question, json_data=None, uid=None):
         """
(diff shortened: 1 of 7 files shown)


yaksh/models.py
```diff
@@ -1284,6 +1284,11 @@ def get_question_answers(self):
                 }]
         return q_a
 
+    def get_latest_answer(self,question_id):
+        all_answers = self.answers.filter(question=question_id)
+        if all_answers:
```
prabhuramachandran: This looks wrong, as I am not sure that all_answers[0] is indeed the latest answer, instead use `self.answers.latest('id')`?

yaksh/test_views.py
```diff
@@ -3378,3 +3379,387 @@ def test_show_statistics_for_student_per_attempt(self):
                                 [1, 1])
         self.assertEqual(response.context['attempts'][0], 1)
         self.assertEqual(response.context['total'], 1)
+
+
+class TestCheck(TestCase):
```
prabhuramachandran: This class name is not nice, can you pick TestQuestionPaper or so?

yaksh/views.py
```diff
@@ -594,11 +597,17 @@ def check(request, q_id, attempt_num=None, questionpaper_id=None):
         else:
             user_answer = request.POST.get('answer')
         if not user_answer:
-            msg = ["Please submit a valid option or code"]
+            msg = "Please submit a valid option or code"
```
prabhuramachandran: Why is this option or code?  Isn't it just an option?

prabhuramachandran: Or maybe option/value?

adityacp: This message is displayed for both code and non-code questions if user answer is empty. So the message is common for both.

yaksh/models.py
```diff
@@ -1284,6 +1284,9 @@ def get_question_answers(self):
                 }]
         return q_a
 
+    def get_latest_answer(self, question_id):
+        return self.answers.filter(question=question_id).order_by("-id").last()
```
prabhuramachandran: This looks wrong, it should be `self.answers.filter(question=question_id).order_by("id").last()`

prabhuramachandran:
My only issue is that the tests did not catch the last answer issue that was seen.  Please add a test for this ASAP.  I may merge this but please add this test.

PR #343, opened 2017-10-05
Fix shuffle questions

adityacp:
This PR fixes shuffled questions bug. 
Now if shuffled questions option is enabled, Answer paper includes shuffled questions irrespective of Question paper questions order.

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 from datetime import datetime, timedelta
 import json
+import random
 import ruamel.yaml
 from ruamel.yaml.scalarstring import PreservedScalarString
 from ruamel.yaml.comments import CommentedMap
@@ -851,6 +852,8 @@ class QuestionPaper(models.Model):
         questions = self.get_ordered_questions()
         for question_set in self.random_questions.all():
             questions += question_set.get_random_questions()
+        if self.shuffle_questions:
+            return self.get_shuffled_questions(questions)
         return questions
 
     def make_answerpaper(self, user, ip, attempt_num):
@@ -941,8 +944,13 @@ class QuestionPaper(models.Model):
             for que_id in que_order:
                 ques.append(self.fixed_questions.get(id=que_id))
         else:
-            ques = self.fixed_questions.all()
+            ques = list(self.fixed_questions.all())
         return ques
+
+    def get_shuffled_questions(self, questions):
+        """Get shuffled questions if auto suffle is enabled"""
+        random.shuffle(questions)
+        return questions
 
     def __str__(self):
         return "Question Paper for " + self.quiz.description
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -924,7 +924,7 @@ class QuestionPaper(models.Model):
     def create_demo_quiz_ppr(self, demo_quiz, user):
         question_paper = QuestionPaper.objects.create(quiz=demo_quiz,
                                                       total_marks=6.0,
-                                                      shuffle_questions=True
+                                                      shuffle_questions=False
                                                       )
         summaries = ['Roots of quadratic equation', 'Print Output',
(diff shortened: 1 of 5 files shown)


yaksh/models.py
```diff
@@ -851,6 +852,8 @@ def _get_questions_for_answerpaper(self):
         questions = self.get_ordered_questions()
         for question_set in self.random_questions.all():
             questions += question_set.get_random_questions()
+        if self.shuffle_questions:
+            return self.get_shuffled_questions(questions)
```
prabhuramachandran: Better to have the return in an else

yaksh/models.py
```diff
@@ -941,9 +947,19 @@ def get_ordered_questions(self):
...
 
+    def get_shuffled_questions(self, questions):
+        """Get shuffled questions if auto suffle is enabled"""
+        random.shuffle(questions)
+        return questions
+
+    def has_questions(self):
+        questions = self.get_ordered_questions() + \
+                    list(self.random_questions.all())
+        return False if len(questions) == 0 else True
```
prabhuramachandran: Just `return len(questions) > 0`

yaksh/models.py
```diff
@@ -1164,13 +1180,27 @@ class AnswerPaper(models.Model):
         default='inprogress'
     )
 
+    # set question order
+    questions_order = models.CharField(max_length=255, blank=True, default='')
```
prabhuramachandran: This is too short a length, what if you have a paper with 100 questions?

yaksh/models.py
```diff
@@ -1164,13 +1180,27 @@ class AnswerPaper(models.Model):
...
+
     objects = AnswerPaperManager()
 
     def current_question(self):
         """Returns the current active question to display."""
         if self.questions_unanswered.all():
-            return self.questions_unanswered.all()[0]
-        return self.questions.all()[0]
+            cur_question = self.get_current_question(
+                self.questions_unanswered.all())
```
prabhuramachandran: You are making two queries here so save the unanswered_questions here and then use that.

yaksh/models.py
```diff
@@ -1164,13 +1180,27 @@ class AnswerPaper(models.Model):
...
+        else:
+            cur_question = self.get_current_question(self.questions.all())
+        return cur_question
+
+    def get_current_question(self, questions):
+        if self.questions_order:
+            question_id = int(self.questions_order.split(',')[0])
```
prabhuramachandran: This logic is not clear.  Why is the first question in the questions_order the current one?

yaksh/models.py
```diff
@@ -1164,13 +1180,27 @@ class AnswerPaper(models.Model):
...
-            return self.questions_unanswered.all()[0]
-        return self.questions.all()[0]
+            cur_question = self.get_current_question(
+                self.questions_unanswered.all())
+        else:
+            cur_question = self.get_current_question(self.questions.all())
+        return cur_question
+
+    def get_current_question(self, questions):
+        if self.questions_order:
```
prabhuramachandran: Why is this if case needed at all, don't you always have a questions_order when you make the paper or is there a case when this doesn't happen?  Maybe there is I am just asking.

yaksh/models.py
prabhuramachandran: Isn't there some other database entry that tells you which is the active question?

yaksh/models.py
adityacp: If this condition is not added then older answer papers where there are no question order will not be able to get questions.

yaksh/models.py
adityacp: No we don't store the current question. we just take the first question from the list of questions.
Then for next question we just iterate over that list based on current question id.

yaksh/models.py
adityacp: I can make it a TextField.

PR #346, opened 2017-10-05
Use MCQ/MCC testcases id instead of options to validate answer

maheshgudi:
Changes the way MCQ, MCC are validated.


prathamesh920:
LGTM

PR #351, opened 2017-10-09
Yaml fixes

### LINKED ISSUE #338: Forcing mandatory fields in yaml file even if empty
When uploading YAML, it assumes that files and tags must be given even if empty which is bad.  I.e. you must have a 
```
files: []
tags: []
```
even if you don't want it.

### LINKED ISSUE #357: Rendering of answers not proper
While viewing user attempts, divs are overlapped.

maheshgudi:
The PR contains the following fixes:

1. Can skip adding file and/or tags field in Yaml question file with test cases. Closes #338 
2. The sidebar on show questions page doesnt overlap on the question list.
3. All questions in view answerpaper accurately display the answer panel heading.  
4. Marks MCC options as  html-safe.
5. Fixes bug where C/C++, Java, and bash wouldn't work with Expected input as None with test case.
6. Fixes user data bug. Closes #357

### DIFF
--- a/yaksh/evaluator_tests/test_bash_evaluation.py
+++ b/yaksh/evaluator_tests/test_bash_evaluation.py
@@ -255,7 +255,6 @@ class BashStdIOEvaluationTestCases(EvaluatorBaseTest):
                              """
                              )
         test_case_data = [{'expected_output': '10',
-                           'expected_input': '',
                            'test_case_type': 'stdiobasedtestcase',
                            'weight': 0.0
                            }]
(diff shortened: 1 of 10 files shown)


maheshgudi:
@prathamesh920 Both **bash and python** work without providing input even with the old code because both bash and python handled the no expected_input condition separately in their respective stdio_evaluators. Java and C didn't. Hence the change.

prathamesh920:
Alright.
But for me C also worked fine.

PR #368, opened 2017-10-12
Fix the Reset Answer javascript and rename it to Clear Text Area

ankitjavalkar:


yaksh/static/yaksh/js/requesthandler.js
```diff
@@ -113,14 +115,17 @@ $(document).ready(function(){
...
 
-  function reset_editor() {
+  $('#code').submit(function(e) {
+    lock_screen();
+    var data = $(this).serializeArray();
+    ajax_check_code($(this).attr("action"), "POST", "html", data, null)
+        e.preventDefault(); // To stop the default form submission.
+  });
+
+  reset_editor = function() {
```
maheshgudi: This will always reset to the init value; which can also be the last submitted code. Are we sure that is what we want? Well from the name clear text area, I was expecting it would clear everything except the snippet.

yaksh/migrations/0008_auto_20171010_0657.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
prathamesh920: I agree that migration file should not have been checked in here.
Also, a reasonable name should be given to the migration file whenever pushed.

yaksh/static/yaksh/js/requesthandler.js
prathamesh920: Yes, the init value is last submitted code if any!
I think reset should simply display snippet on click or blank if no snippet.

yaksh/migrations/0008_auto_20171010_0657.py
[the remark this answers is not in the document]
ankitjavalkar: Needs to be removed, agreed

yaksh/static/yaksh/js/requesthandler.js
ankitjavalkar: Will rename button to 'Undo Changes'

yaksh/templates/yaksh/question.html
```diff
@@ -204,7 +204,7 @@
                              <h4>Write your program below:</h4>
                          </div>
                          <div class="col-md-3">
-                             <a href="#" class="pull-right" onclick="reset_editor()" name="reset" id="reset">Reset Answer <span class="glyphicon glyphicon-refresh"></span></a>
+                             <a href="#answer" class="pull-right" onclick="reset_editor()" name="reset" id="reset">Undo Changes<span class="glyphicon glyphicon-refresh"></span></a>
```
maheshgudi: A suggestion, could you add a space between Undo Changes and refresh glyphicon.

yaksh/static/yaksh/js/requesthandler.js
```diff
@@ -83,6 +83,8 @@ function ajax_check_code(url, method_type, data_type, data, uid) {
 }
 
 var global_editor = {};
+var reset_editor;
```
maheshgudi: Is the variable declaration necessary?

ankitjavalkar:
Fixed as per suggestions

PR #373, opened 2017-10-26
Detailed csv download quiz

### LINKED ISSUE #354: Dump detailed quiz attempts to csv
CSV dump should have the following details:
- course/quiz/attempt number
- user details
- individual question details.
Current CSV does not have details about questions and
only users who took the quiz, only their data is in CSV dump.
This is slightly inconsistent and inconvenient when comparing
attempts via CSV dump.
So, if an user in a course has not attempted the quiz 
then in CSV dump zero marks will be allotted against his name.

prathamesh920:
CSV download for a quiz now shows question wise grades.
Also, for a given attempt all the users from the course are entered in
the CSV. If the user has not attempted then a dash '-' is put under the
grades.
Also, handles random questions, if a question paper has questions
selected from pool of questions then all the questions are entered in
the CSV. 'NA' is put under the question grade if that question has not
come in the question/answer paper for that given user
Closes #354


yaksh/views.py
```diff
@@ -1264,49 +1273,79 @@ def user_data(request, user_id, questionpaper_id=None):
...
+def download_quiz_csv(request, course_id, quiz_id):
+    current_user = request.user
+    if not is_moderator(current_user):
         raise Http404('You are not allowed to view this page!')
-    quiz = Quiz.objects.get(questionpaper=questionpaper_id)
+    course = get_object_or_404(Course, id=course_id)
+    quiz = get_object_or_404(Quiz, id=quiz_id)
+    if not course.is_creator(current_user) and not course.is_teacher(current_user):
+        raise Http404('The quiz does not belong to your course')
+    users = course.get_enrolled().order_by('first_name')
```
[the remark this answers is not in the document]
prathamesh920: Alright!
For now below in the views I have added exclude call on the users to remove teachers and creator.

yaksh/templates/yaksh/monitor.html
```diff
@@ -115,4 +118,43 @@
...
+    <div class="modal-body">
+        {% for field in csv_fields %}
+        <div class="form-check form-check-inline">
+          <label class="form-check-label">
+              <input class="form-check-input" name="csv_fields" type="checkbox" value="{{ field }}" checked> {{ field }}
+          </label>
+        </div>
+        {% endfor %}
+        <b>Select Attempt Number: Default latest attempt</b>
+        <select class="form-control" id = "attempt_number">
```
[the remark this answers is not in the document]
prathamesh920: Yes correct there should be a name attribute! Yes, in views I was setting to the latest attempt always.
Fixed now.

yaksh/templates/yaksh/monitor.html
```diff
@@ -115,4 +118,46 @@
...
+
+<!-- CSV Modal -->
+<div class="modal fade" id="csvModal" role="dialog">
+    <div class="modal-dialog">
+
+    <!-- Modal content-->
+    <div class="modal-content">
+    <div class="modal-header">
+      <button type="button" class="close" data-dismiss="modal">&times;</button>
+      <h3 class="modal-title">Uncheck unwanted columns</h3>
```
adityacp: Maybe you can change this to `Tick\Untick for specific column`.

yaksh/models.py
```diff
@@ -840,6 +840,13 @@ class QuestionPaper(models.Model):
 
     objects = QuestionPaperManager()
 
+    def get_question_bank(self):
```
adityacp: Just a suggestion, we already have a function to get all questions for a question paper.
You can try that instead of this extra function.

yaksh/test_models.py
```diff
@@ -384,6 +384,26 @@ def setUpClass(self):
         self.questions = Question.objects.filter(active=True)
         self.quiz = Quiz.objects.get(description="demo quiz 1")
 
+        # create question paper with only fixed questions
+        self.question_paper_fixed_questions = QuestionPaper.objects.create(
+                quiz=self.quiz)
+        self.question_paper_fixed_questions.fixed_questions.add(
+                self.questions.get(id=11), self.questions.get(id=10))
```
adityacp: It will be better if you fetch question with some other parameter instead of id. This id might change at some point.

yaksh/test_models.py
```diff
@@ -384,6 +384,26 @@ def setUpClass(self):
...
+
+        # create question paper with only random questions
+        self.question_paper_random_questions = QuestionPaper.objects.create(
+                quiz=self.quiz)
+        self.question_set_random = QuestionSet.objects.create(marks=2,
+            num_questions=2)
+        self.question_set_random.questions.add(self.questions.get(id=13),
+                self.questions.get(id=5), self.questions.get(id=7))
```
adityacp: Same as above.

yaksh/test_models.py
```diff
@@ -441,6 +461,31 @@ def setUpClass(self):
         self.trial_course = Course.objects.create_trial_course(self.user)
         self.trial_quiz = Quiz.objects.create_trial_quiz(self.trial_course, self.user)
 
+
+    def test_get_question_bank(self):
+        # Given
+        ids = [11, 10]
+        questions = list(Question.objects.filter(id__in=ids))
```
adityacp: Same again

yaksh/views.py
```diff
@@ -1264,49 +1273,80 @@ def user_data(request, user_id, questionpaper_id=None):
...
+    csv_fields = []
+    attempt_number = None
+    question_paper = quiz.questionpaper_set.last()
+    last_attempt_number =AnswerPaper.objects.get_attempt_numbers(question_paper.id).last()
+    if request.method == 'POST':
+        csv_fields = request.POST.getlist('csv_fields')
+        attempt_number = request.POST.get('attempt_number', last_attempt_number)
+    if not csv_fields:
+        csv_fields = ['name', 'username', 'roll_number', 'institute',
+                'department', 'questions', 'total', 'out_of', 'percentage', 'status']
```
adityacp: Just a suggestion instead of `total` can we write it as `marks_obtained`?

yaksh/test_models.py
ankitjavalkar: Creating specific questions for this particular test case would be a good idea.

yaksh/views.py
```diff
@@ -1007,11 +1012,15 @@ def monitor(request, quiz_id=None):
                     attempt_number=last_attempt['last_attempt_num']
                 )
             )
+    csv_fields = ['name', 'username', 'roll_number', 'institute',
+            'department', 'questions', 'total', 'out_of', 'percentage', 'status']
```
ankitjavalkar: Can we store the CSV headings ```csv_fields``` in a global variable in ```models.py``` similar to how we store ```question_types``` and ```languages```, this gives it a central place to change instead of hiding it in a view function. Please suggest or feel free to ignore.

yaksh/templates/yaksh/monitor.html
prathamesh920: Changed the text.

yaksh/views.py
prathamesh920: Changed

yaksh/test_models.py
prathamesh920: Okay then used summary instead of id.
@ankitjavalkar Actually I only need to show that the required questions are retrieved.
So questions can be any from the present set. Then too any suggestion then please mention.

yaksh/test_models.py
prathamesh920: Used summary

yaksh/views.py
prathamesh920: I too thought of giving it a central place. But models is not the place because in models ```question_types``` and ```languages``` are used by our models, but csv_fields have no relation with models for now so placing there will not be proper.
For now keeping a variable CSV_FIELDS in views itself as the central place.

PR #375, opened 2017-10-27
Edit Profile Bug Fix

### LINKED ISSUE #374: Edit Profile Bug
Decorator has_profile will redirect to edit_profile if no profile.
But then if I submit my profile form then the has_profile will again
redirect to edit_profile, instead of updating my profile.
So this cycle will continue endlessly for a user with no profile, and
will never be able to create/update his profile!
Will face this when user is created via csv upload, django admin or  oauth login without pipeline.
Also, profile instance is passed to the profile form via get query, which will fail if no profile.

prathamesh920:
edit_profile view had a decorator has_profile.
So, has_profile will redirect to edit_profile if no profile.
But then if I submit my profile form then the has_profile will again
redirect to edit_profile, instead of updating my profile.
So this cycle will continue endlessly for a user with no profile, and
will never be able to create/update his profile!

Will face this when user is created via csv upload, django admin or
oauth login without pipeline.

Also, profile instance is passed to the profile form via get query,
which will fail if no profile.

Added a views test for the above.
Fixed. Closes #374 

Additionally that can be thought of later:
The has_profile decorator is used for few views only,
so one can access views if they know the url, even if they do not have a
profile.
email edit option for users

### DIFF
--- a/yaksh/test_views.py
+++ b/yaksh/test_views.py
@@ -196,6 +196,38 @@ class TestProfile(TestCase):
         self.assertEqual(response.status_code, 200)
         self.assertTemplateUsed(response, 'yaksh/profile_updated.html')
 
+    def test_edit_profile_post_for_user_without_profile(self):
+        """
+        POST request to edit_profile view should update the user's profile
+        """
+        self.client.login(
+            username=self.user1.username,
+            password=self.user1_plaintext_pass
+        )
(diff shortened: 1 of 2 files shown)


ankitjavalkar:
Please add a line to the CHANGELOG. All else LGTM

PR #378, opened 2017-11-06
Upload users csv

prathamesh920:
Upload user to the course via csv using UI

The csv takes firstname, lastname and email.
User and Profile are created with username and password been same as
email.
Following cases are handled for csv upload:
- wrong csv headers
- missing csv values
- already existing users
- invalid csv


yaksh/file_utils.py
```diff
@@ -50,3 +50,30 @@ def extract_files(zip_file, path=None):
...
+
+def is_csv(document):
+    try:
+        try:
+            content = document.read(1024).decode('utf-8')
+        except AttributeError:
+            document.seek(0)
+            content = document.read(1024)
+        sniffer = csv.Sniffer()
+        dialect = sniffer.sniff(content)
```
adityacp: Is this `dialect` variable used anywhere else? If not then it is unnecessary.

yaksh/file_utils.py
```diff
@@ -50,3 +50,30 @@ def extract_files(zip_file, path=None):
...
+        document.seek(0)
+    except (csv.Error, UnicodeDecodeError):
+        return False
+    return True
```
adityacp: Cant we just add `except AttributeError` as different exception instead of nesting. 
For Ex. `try then except UnicodeDecodeError except AttributeError`

yaksh/test_views.py
```diff
@@ -1758,6 +1760,141 @@ def tearDown(self):
...
         self.user1_course.delete()
 
+
+    def test_upload_users_with_correct_csv(self):
+        # Given
+        self.client.login(
+            username=self.user1.username,
+            password=self.user1_plaintext_pass
+        )
+        csv_file_path = os.path.join(settings.FIXTURE_DIRS, "users_correct.csv")
```
ankitjavalkar: please avoid using FIXTURE_DIRS as the PR #379 deprecates this value in ```settings.py``` and movese it to ```models.py``` as FIXTURES_DIR_PATH
Please see https://github.com/ankitjavalkar/online_test/blob/3ff5eeeb1066d579213c9ca7b62c1e39d774166c/yaksh/models.py#L85

yaksh/views.py
```diff
@@ -1796,6 +1796,69 @@ def download_assignment_file(request, quiz_id, question_id=None, user_id=None):
...
+        if 'csv_file' not in request.FILES:
+            context['message'] = "Please upload a CSV file."
+            return my_render_to_response('yaksh/course_detail.html', context,
+                    context_instance=ci)
+        csv_file = request.FILES['csv_file']
+        if not is_csv(csv_file):
+            context['message'] = "The file uploaded is not a CSV file."
+            return my_render_to_response('yaksh/course_detail.html', context,
+                    context_instance=ci)
+        headers = {'firstname':'', 'lastname':'', 'email':''}
```
adityacp: Missing space after semicolon in `headers` dictionary.

yaksh/views.py
```diff
@@ -1796,6 +1796,69 @@ def download_assignment_file(request, quiz_id, question_id=None, user_id=None):
...
+                    counter, email))
+            else:
+                Profile.objects.create(user=user, is_email_verified=True)
+                course.students.add(user)
+                upload_details.append("{0} -- {1} -- User Added Successfully".format(
+                    counter, email))
+    if counter == 0:
+        upload_details.append("No rows in the CSV file")
+    return upload_details
+
```
adityacp: Can you please check the code with pep8? This will reduce the pep8 formatting later.

yaksh/views.py
```diff
@@ -1796,6 +1796,69 @@ def download_assignment_file(request, quiz_id, question_id=None, user_id=None):
...
+            continue
+        user = User.objects.filter(email=email)
+        if user.exists():
+            upload_details.append("{0} -- {1} -- Email Already Exists".format(
+                counter, email))
+        else:
+            try:
+                user = User.objects.create_user(username=email, password=email,
+                        email=email, first_name=first_name, last_name=last_name)
+            except IntegrityError:
```
adityacp: Just to be on safe side you can do django import as `from django.db import IntegrityError`

yaksh/views.py
```diff
@@ -1796,6 +1796,69 @@ def download_assignment_file(request, quiz_id, question_id=None, user_id=None):
...
+                upload_details.append("{0} -- {1} -- User Already Exists".format(
+                    counter, email))
+            else:
+                Profile.objects.create(user=user, is_email_verified=True)
+                course.students.add(user)
```
adityacp: I feel it would be better to add the users to the course outside for loop. You can pass the users list as course.students.add(*users)

yaksh/views.py
```diff
@@ -1796,6 +1796,69 @@ def download_assignment_file(request, quiz_id, question_id=None, user_id=None):
...
+            context['message'] = "Please upload a CSV file."
+            return my_render_to_response('yaksh/course_detail.html', context,
+                    context_instance=ci)
+        csv_file = request.FILES['csv_file']
+        if not is_csv(csv_file):
+            context['message'] = "The file uploaded is not a CSV file."
+            return my_render_to_response('yaksh/course_detail.html', context,
+                    context_instance=ci)
+        headers = {'firstname':'', 'lastname':'', 'email':''}
+        reader = csv.DictReader(csv_file.read().decode('utf-8').splitlines())
```
maheshgudi: This piques my interest!! Why does the decode thing not raise an error in python 3?

yaksh/views.py
maheshgudi: We literally impose no checks on the email id format,. We can have a regex validator, to check if the email is legit or not

yaksh/file_utils.py
prathamesh920: Actually dialect can be used while reading csv but I am not using so will remove it.

yaksh/file_utils.py
prathamesh920: I had tried it as you can see in my commits, but then does not fulfill the requirement so kept nested.
Though can remove nested but will have two separate try..except.

yaksh/test_views.py
prathamesh920: Yes, the PR #379 is been merged so will fix it!

yaksh/views.py
prathamesh920: @adityacp , I agree, this will optimize the code for large data, but still need to check completely.
Changes done.

@maheshgudi 
Not validated email id because even if we validate with regex, there can be email ids with correct structure but in reality they do not exists . We are providing update email option and also I am setting email_verified True as these are been uploaded by the authenticated user that is moderator, I feel its fine if we do not validate for now.

yaksh/views.py
prathamesh920: I agree, your suggestion is better.
Done

yaksh/views.py
prathamesh920: Done

yaksh/views.py
prathamesh920: Done

yaksh/file_utils.py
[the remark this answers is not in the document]
prathamesh920: Explicitly defining delimiter will limit this util function.
By default I was using 'comma' as the delimiter.
So to handle any delimiters, I will return the dialect and pass it in the csv read to identify the delimiter function.

yaksh/templates/yaksh/course_detail.html
```diff
@@ -41,6 +41,19 @@
         <a href="{{URL_ROOT}}/exam/manage/send_mail/{{ course.id }}/">
         Send Mail</a>
         </li>
+        <li>
+            <form id="upload_users" action="{{ URL_ROOT }}/exam/manage/upload_users/{{course.id}}/"
+                method="POST" enctype="multipart/form-data">
+                {% csrf_token %}
+                <input type="file" name="csv_file" />
+                <button class="btn btn-primary" type=submit> Upload Users <span class="glyphicon glyphicon-open"/></button>
```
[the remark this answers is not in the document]
prathamesh920: Why break line!
I am using a div with particular class after button so no need to put br. Next element is already on next line.
Even if I put a break line it will show as it is.

yaksh/views.py
```diff
@@ -1796,6 +1796,69 @@ def download_assignment_file(request, quiz_id, question_id=None, user_id=None):
 
 @login_required
 @email_verified
+def upload_users(request, course_id):
```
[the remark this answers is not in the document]
prathamesh920: Because its a django request.
And course details have send mail, clone course, enroll, reject etc. all are separate django views,
none are in course details view!
And no harm in having a separate view.

prathamesh920:
@maheshgudi 
Not validated email id because even if we validate with regex, there can be email ids with correct structure but in reality they do not exists . We are providing update email option and also I am setting email_verified True as these are been uploaded by the authenticated user that is moderator, I feel its fine if we do not validate for now.

yaksh/templates/yaksh/course_detail.html
```diff
@@ -41,6 +41,19 @@
         <a href="{{URL_ROOT}}/exam/manage/send_mail/{{ course.id }}/">
         Send Mail</a>
         </li>
+        <li>
+            <form id="upload_users" action="{{ URL_ROOT }}/exam/manage/upload_users/{{course.id}}/"
```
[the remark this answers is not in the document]
prathamesh920: Internally its form, visually for end user its just an upload button.
I agree that it should have been better UI, but the current course details page is such that it looks hanging on the page with other elements.

yaksh/file_utils.py
```diff
@@ -50,3 +50,30 @@ def extract_files(zip_file, path=None):
...
+
+
+def headers_present(dict_reader, headers):
+    fields = dict_reader.fieldnames
```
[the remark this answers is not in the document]
prathamesh920: Yes

yaksh/file_utils.py
```diff
@@ -50,3 +50,30 @@ def extract_files(zip_file, path=None):
...
+    header_fields = set()
+    for field in fields:
```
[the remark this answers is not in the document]
prathamesh920: I think I am using sniffer, so won't blow up.
I think developer using the util function should take care of the arguments

yaksh/file_utils.py
```diff
@@ -50,3 +50,30 @@ def extract_files(zip_file, path=None):
...
+        if field.strip() in headers.keys():
+            headers[field.strip()] = field
+            header_fields.add(field.strip())
+    if  header_fields != set(headers.keys()):
```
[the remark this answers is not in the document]
prathamesh920: changing the function.

yaksh/file_utils.py
```diff
@@ -50,3 +50,30 @@ def extract_files(zip_file, path=None):
...
+        return False
+    return True
```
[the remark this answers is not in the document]
prathamesh920: changing the function.

yaksh/views.py
```diff
@@ -1843,6 +1843,145 @@ def download_assignment_file(request, quiz_id, question_id=None, user_id=None):
...
+    remove = "false"
+    email, first_name, last_name = map(str.strip, [row['email'],
+                                       row['firstname'],
+                                       row['lastname']])
+    if 'password' in fields:
+        password = row['password'].strip()
+        if not password:
+            password = email
+    else:
+        password = email
```
adityacp: I feel password = email is redundant in both if and else. You can do password = email by default and check if the password exists in csv if it exists then assign that value to password variable i.e. password = row['password'].strip()

yaksh/views.py
```diff
@@ -1843,6 +1843,145 @@ def download_assignment_file(request, quiz_id, question_id=None, user_id=None):
...
+    if 'institute' in fields:
+        institute = row['institute'].strip()
+    if 'department' in fields:
+        department = row['department'].strip()
+    if 'remove' in fields:
+        remove = row['remove'].strip()
+    if 'username' in fields:
+        username = row['username'].strip()
+    else:
+        username = email
```
adityacp: Same as above

yaksh/views.py
```diff
@@ -1843,6 +1843,145 @@ def download_assignment_file(request, quiz_id, question_id=None, user_id=None):
...
+        profile_defaults = {'institute': institute, 'roll_number': roll_no,
+                            'department': department, 'is_email_verified': True}
+        _create_or_update_profile(user, profile_defaults)
+        if created:
+            state = "Added"
+            user.set_password(password)
+            user.save()
+            course.students.add(user)
+        else:
+            state = "Updated"
```
adityacp: while updating if I change password field for a user the password is not updated. I guess you need to add below lines outside if and else 
<pre>user.set_password(password) 
user.save()</pre>

prathamesh920: Yes, initially thought should not allow update password by the moderator.
But, yes this is required, incase the user changes the password and forget, then moderator can update!

prathamesh920:
I may shift few of the functions from views to models, wherever feasible.

yaksh/templates/yaksh/course_detail.html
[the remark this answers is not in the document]
prathamesh920: I have currently added form on the top of the page, removed from sidebar, toggle thing I am not doing currently because the upload details are returned to the main course detail page. So it will appear as if upload and upload details are on different page.

yaksh/templates/yaksh/course_detail.html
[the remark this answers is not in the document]
prathamesh920: Exactly, intentions are not to separate out because the button and instructions look one piece.
So, user will definitely will read the instructions. You can once see UI wise, I am uploading the image. 
If not good then will add break.

prathamesh920:
user_upload

yaksh/views.py
```diff
@@ -1843,6 +1843,145 @@ def download_assignment_file(request, quiz_id, question_id=None, user_id=None):
...
+                                             context_instance=ci)
+        reader.fieldnames = stripped_fieldnames
+        context['upload_details'] = _read_user_csv(reader, course)
+    return my_render_to_response('yaksh/course_detail.html', context,
+                                 context_instance=ci)
+
+
+def _read_user_csv(reader, course):
+    fields = reader.fieldnames
+    upload_details = ["Upload Summary:"]
```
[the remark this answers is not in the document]
prathamesh920: Just a heading along with upload details that will be seen to the moderator after upload.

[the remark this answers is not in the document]
prathamesh920: I have currently added form on the top of the page, removed from sidebar, toggle thing I am not doing currently because the upload details are returned to the main course detail page. So it will appear as if upload and upload details are on different page.



Exactly, intentions are not to separate out because the button and instructions look one piece.
So, user will definitely will read the instructions. You can once see UI wise, I am uploading the image.
If not good then will add break.

PR #389, opened 2017-11-14
Multiple minor fixes (Listed in description)

ankitjavalkar:
- Fix README instruct to run docker container before running django server
- Add messages after commands execute succesfully
- Touch a new file to denote successful running of code server

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import invoke
 from invoke import task
 import os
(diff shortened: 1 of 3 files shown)


tasks.py
```diff
@@ -6,6 +7,9 @@
 SCRIPT_DIR = os.path.abspath(os.path.dirname(__file__))
 TARGET_CONTAINER_NAME = 'yaksh_code_server'
 SRC_IMAGE_NAME = 'fossee/yaksh_codeserver'
+CHECK_FILE = 'touch_me_not.txt'
```
prabhuramachandran: Maybe this can be be 'server_running.txt'

tasks.py
```diff
@@ -84,12 +89,18 @@ def start(ctx, ports=SERVER_POOL_PORT, image=SRC_IMAGE_NAME, unsafe=False,
...
+
+
 @task
 def stop(ctx, container=TARGET_CONTAINER_NAME, hide=True):
     result = ctx.run("sudo docker ps -q --filter='name={0}'".format(container))
     if result.stdout:
         print ("** Discarding the docker container <{0}>".format(container))
         ctx.run("sudo docker stop {0}".format(container))
         ctx.run("sudo docker rm {0}".format(container))
+        print ("** Done! Discarded the docker container <{0}>".format(container))
```
prabhuramachandran: You should also remove the CHECK_FILE on stop otherwise it will always find the fine even the next time you run the container.

prabhuramachandran: And you should remove it before you run the docker command in case that fails.

tasks.py
ankitjavalkar: Fixed

tasks.py
ankitjavalkar: Fixed

prabhuramachandran:
LGTM, except a migration seems to have snuck in.

ankitjavalkar:
Apologies, reset the complete commit and all the changes are now in a single commit without the migrations

PR #400, opened 2017-11-22
Display error output without page reloading

### LINKED ISSUE #399: Display error without question page reloading
On submitting an answer to a question during a quiz, the page reloads. This shouldn't happen.

maheshgudi:
- On submitting an answer, the error output were displayed after reloading the page. Fixes this.
- Closes #399


yaksh/templates/yaksh/error_template.html
```diff
@@ -0,0 +1,93 @@
+{% block css%}
+    <link rel="stylesheet" href="{{ URL_ROOT }}/static/yaksh/css/dashboard.css" type="text/css" />
+{% endblock %}
+
+{% load custom_filters %}
+
+{% if error_message %}
+<div class="col-md-12">
```
adityacp: Can make this class name as `col-md-10 col-md-offset-2 main` so as to avoid overlapping of side bar.

yaksh/views.py
```diff
@@ -653,13 +653,33 @@ def check(request, q_id, attempt_num=None, questionpaper_id=None):
 @csrf_exempt
 def get_result(request, uid):
     result = {}
+    template_dir = os.path.dirname(os.path.realpath(__file__))
+    template_path = os.path.join(*[template_dir, 'templates',
+                                   'yaksh','error_messages.html'
+                                   ])
```
adityacp: Is this needed? Since I don't see error_messages.html is added.

yaksh/static/yaksh/js/requesthandler.js
```diff
@@ -53,16 +53,32 @@ function response_handler(method_type, content_type, data, uid){
...
         }
-        check_state(request_status, uid);
     } else {
         reset_values();
         unlock_screen();
     }
 }
 
+function focus_on_error(){
+    var ele = document.getElementById('error_panel')
```
adityacp: Just a suggestion you don't need to get element again you can pass the element as an argument to this function from response_handler.

PR #406, opened 2017-12-18
Fix assignment upload to work with new code server

### LINKED ISSUE #388: Assignment Upload with new code server
Assignment can be uploaded, but cannot be checked instantaneously during the upload.

maheshgudi:
1. Fixes assignment upload to work the new tornado server. 
2. Closes #388

### DIFF
--- a/yaksh/static/yaksh/js/requesthandler.js
+++ b/yaksh/static/yaksh/js/requesthandler.js
@@ -81,31 +81,47 @@ function focus_on_error(ele){
       window.scrollBy(0, -15);
         }
     }
+function csrfSafeMethod(method) {
+    // these HTTP methods do not require CSRF protection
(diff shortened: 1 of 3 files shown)


yaksh/static/yaksh/js/requesthandler.js
```diff
@@ -126,19 +142,26 @@ $(document).ready(function(){
...
-
+}
+if (question_type == 'upload' || question_type == 'code') {
   $('#code').submit(function(e) {
     lock_screen();
+    if (question_type == "code"){
     var data = $(this).serializeArray();
+  }
+  else if (question_type == "upload"){
+    var data = new FormData($("#code").get(0));
```
prathamesh920: Just a suggestion.
Instead of get(),  we can pass form using document.getElementById, just for clarity.

PR #408, opened 2017-12-27
Exercise feature in video lessons

### LINKED ISSUE #386: Exercises in the learning unit
Can have exercises for practice.
Weightage can be zero for the exercise questions.

prathamesh920:
Exercise is same as quiz except for following differences:
- no time limit
- no marks weightage
- no instruction page
- skip denied for a particular time

An attribute 'is_exercise' in Quiz determines whether the quiz is
exercise or not.
Questions contains 'min_time' attribute.
For an exercise a question cannot be skipped for an allotted minimum
time, after which either django or JavaScript makes Next button
available.

Implementation is as such due to complexity of our existing views and
templates.
Also, after min_time, same question with Next button is available to
move on, assuming that solution is present in the video.

Any better suggestion then kindly suggest.

Closes #386


yaksh/views.py
```diff
@@ -326,6 +326,56 @@ def add_quiz(request, quiz_id=None, course_id=None):
...
+        if form.is_valid():
+            if quiz is None:
+                form.instance.creator = user
+            quiz = form.save(commit=False)
+            quiz.is_exercise = True
+            quiz.time_between_attempts = 0
+            quiz.weightage = 0
+            quiz.allow_skip = False
+            quiz.attempts_allowed = -1
+            quiz.duration = 1000
```
maheshgudi: Problem with this is no exercise is going to work after 1000 mins (or appx 17 hrs). This is a bad solution, since exercises should inherently be timeless.

yaksh/views.py
```diff
@@ -505,27 +556,42 @@ def start(request, questionpaper_id=None, attempt_num=None, course_id=None,
...
 @email_verified
 def show_question(request, question, paper, error_message=None, notification=None,
-                  course_id=None, module_id=None):
+                  course_id=None, module_id=None, previous_question=None):
     """Show a question if possible."""
     user = request.user
+    quiz = paper.question_paper.quiz
+    quiz_type = 'Exam'
+    can_skip = False
+    if previous_question:
```
maheshgudi: Same with timed questions. If the ```duration``` for exercise is set to a time smaller than the question duration, some how clicking on the submit button does nothing.

yaksh/views.py
```diff
@@ -731,9 +814,12 @@ def get_result(request, uid, course_id, module_id):
         next_question, error_message, paper = _update_paper(request, uid,
                                                             result
                                                             )
+        answer = Answer.objects.get(id=uid)
+        current_question = answer.question
```
maheshgudi: Food for thought, if I have a question with time still left, and I give the correct answer, shouldn't I be able to go to the next question?

yaksh/views.py
maheshgudi: Also for God mode, the moderator should be able to go to the next question, irrespective of the question duration.

yaksh/models.py
```diff
@@ -747,6 +749,8 @@ class Question(models.Model):
     # Check assignment upload based question
     grade_assignment_upload = models.BooleanField(default=False)
 
+    min_time =  models.IntegerField("time in minutes", default=0)
```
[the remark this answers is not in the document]
prathamesh920: Yes but did not add sample answer yet because the solution will be in the video, and there can be multiple answers for the question.
Will implement this locally and will push after a brief discussion.

yaksh/views.py
prathamesh920: Yes I agree not a good solution, but this is just to replace database default value.
But this will work timeless as in views there is no time check.
Also if the user starts the exercise and is inactive for 17 hours the exercise may end, but then we have infinite attempts allowed as it is an exercise.

God mode and user mode will work as our quiz works because while cloning is_exercise is False.
This needs to be True for User mode. User mode and God mode are currently disabled for Exercise.  Will enable it.

yaksh/views.py
prathamesh920: This works. Actually submit button won't go to the next question.
User has to click on Next to visit next question.

yaksh/views.py
```diff
@@ -612,7 +687,8 @@ def check(request, q_id, attempt_num=None, questionpaper_id=None,
                 msg = "Please enter an Integer Value"
                 return show_question(
                     request, current_question, paper, notification=msg,
-                    course_id=course_id, module_id=module_id
+                    course_id=course_id, module_id=module_id,
```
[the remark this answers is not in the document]
prathamesh920: Actually in start view function we have check for this.
Will check this

yaksh/views.py
prathamesh920: Considering that we have solution in the video so the user should wait till the question time is over.

yaksh/views.py
```diff
@@ -510,27 +561,42 @@ def start(request, questionpaper_id=None, attempt_num=None, course_id=None,
...
         return complete(
             request, msg, paper.attempt_number, paper.question_paper.id,
             course_id=course_id, module_id=module_id
         )
-    if not paper.question_paper.quiz.active:
+    if not quiz.active:
         reason = 'The quiz has been deactivated!'
         return complete(
             request, reason, paper.attempt_number, paper.question_paper.id,
             module_id=module_id
```
adityacp: This will fail once the attempt is started and the quiz is deactivated. You need to send `course_id` as keyword arg to complete function.

yaksh/templates/yaksh/courses.html
```diff
@@ -151,9 +151,14 @@
                     {% for unit in module.get_learning_units %}
                         <ul class="inputs-list">
                         <li>
-                        {% if unit.type == "quiz" %}
-                            <a href="{{URL_ROOT}}/exam/manage/addquiz/{{unit.quiz.id}}/{{course.id}}">
-                            {{unit.quiz.description}}</a>
+                        {% if unit.learning_type == "quiz" %}
```
adityacp: This should `type` instead of `learning_type`

yaksh/templates/yaksh/courses.html
```diff
@@ -151,9 +151,14 @@
...
+                            {% if unit.quiz.is_exercise %}
```
adityacp: You need to add this condition below in allotted courses. Since teacher can also view courses. Currently allotted courses and created courses are displayed separately.

yaksh/templates/yaksh/question.html
```diff
@@ -184,6 +197,23 @@ <h4>Files to download for this question</h4>
                       <h5><a href="{{f_name.file.url}}">{{f_name.file.name}}</a></h5>
                       {% endfor %}
                       {% endif %}
+
+                        {% if quiz.is_exercise %}
+                            {% if question.solution %}
+                                <h4><u> Solution by teacher</u></h4>
```
adityacp: Solution by the teacher heading is shown by default, I feel it should be activated if the student gets the right answer or after the timeout.

yaksh/forms.py
```diff
@@ -173,6 +173,12 @@ def clean(self):
         return user
 
 
+class ExerciseForm(forms.ModelForm):
+    class Meta:
+        model = Quiz
+        fields = ['description']
```
adityacp: I feel Exercise form should also have `allow students to view answer paper` if suppose students want to view their answer papers.

yaksh/templates/yaksh/courses.html
ankitjavalkar: Wasn't this supposed to be changed to ```category``` ?

adityacp: I changed this to `type`. I feel it is ok.

adityacp:
After quitting from exercise quit template shows following message:-

1. `Are you sure you wish to quit the exam?`
   - I feel instead of `exam` it can be `exercise` since exercise is not actual exam.
2. `Be sure, as you won't be able to restart this exam.`
   - This should not come since exercise has infinite attempts.

adityacp:
Can you please change `course.is_active_enrollment` to `course.is_active_enrollment()` in the following lines mentioned below:-
1. https://github.com/prathamesh920/online_test/blob/44e16651a9f7faec95a32c19738a996a731ae57a/yaksh/views.py#L958

2. https://github.com/prathamesh920/online_test/blob/44e16651a9f7faec95a32c19738a996a731ae57a/yaksh/views.py#L1031

yaksh/views.py
prathamesh920: I have not touched this part!
But since it is an error I will check this, as there are other similar function call!

prathamesh920:
Yes sure

yaksh/templates/yaksh/courses.html
prathamesh920: done

yaksh/templates/yaksh/courses.html
prathamesh920: Added

yaksh/templates/yaksh/question.html
prathamesh920: Done.
Actually causes slight redundancy so implemented like this.

yaksh/forms.py
prathamesh920: Okay providing the option

yaksh/views.py
adityacp: I have checked this is the only function which is missed. But still, you can check

PR #414, opened 2018-01-08
Fix Oauth email verification

adityacp:
This PR fixes the bug where email verified is not set to True if the user logs in using OAuth system.

**Not Mergeable for now**
Need to test on the server.

Edit:
Tested on ```yaksh.fossee.in```


ankitjavalkar:
Tested this on ```yaksh.fossee.in```, User is able to login, Profile is created with the ```is_email_verified``` set to True.
LGTM

maheshgudi:
Works for me too. Can be merged.

PR #422, opened 2018-01-12
Show course status

### LINKED ISSUE #421: Show Course Status
A moderator should be able to view student progress for a particular course.

adityacp:
This PR allows moderator or teacher to view student progress for a particular course.
Closes #421

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -1,4 +1,5 @@
 from __future__ import unicode_literals
+from __future__ import division
 from datetime import datetime, timedelta
 import json
 import random
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -453,12 +453,16 @@ class LearningModule(models.Model):
         ordered_units = learning_module.learning_unit.order_by("order")
         status_list = [unit.get_completion_status(user, course)
                        for unit in ordered_units]
-        if all([status == "completed" for status in status_list]):
-            return "completed"
+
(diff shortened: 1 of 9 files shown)


yaksh/models.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import unicode_literals
+from __future__ import division
```
maheshgudi: Unicode literals and division can be called together

yaksh/views.py
```diff
@@ -2666,3 +2666,20 @@ def course_modules(request, course_id, msg=None):
     context = {"course": course, "learning_modules": learning_modules,
                "user": user, "msg": msg}
     return my_render_to_response('yaksh/course_modules.html', context)
+
+
+@login_required
+@email_verified
+def course_status(request, course_id):
+    user = request.user
```
maheshgudi: Instead of the export to CSV through js, I suggest it would be good if another argument, a Boolean one, is passed. If the Boolean is false it will render in HTML or else it will create a CSV file and dump the data.

adityacp: I felt it would be better to do because we can avoid request and unnecessary db query which would result in same data fetch which I am already fetching.

maheshgudi: Yes, but my way does the same thing, without the intereference of js

adityacp: I feel it is okay for now

PR #429, opened 2018-01-20
Shuffle testcases

### LINKED ISSUE #332: Shuffle the order of MCQ, MCC options
Options in MCQ and MCC questions should be shuffled for each student.

maheshgudi:
- Each user will have separate test cases/options for MCQ/MCC question if shuffle test case options is checked.
- Closes #332

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -936,13 +936,13 @@ class Question(models.Model):
 
     def get_ordered_test_cases(self, answerpaper):
         try:
-            order = TestcaseOrder.objects.get(answer_paper=answerpaper,
+            order = TestCaseOrder.objects.get(answer_paper=answerpaper,
                                               question = self
-                                              ).testcase_order.split(",")
+                                              ).order.split(",")
             return [self.get_test_cases(id=int(tc_id))[0]\
                     for tc_id in order
                     ]
-        except TestcaseOrder.DoesNotExist:
+        except TestCaseOrder.DoesNotExist:
             return self.get_test_cases()
 
     def get_maximum_test_case_weight(self, **kwargs):
@@ -1221,10 +1221,10 @@ class QuestionPaper(models.Model):
                 if question.shuffle_testcases:
                     random.shuffle(testcases)
                 testcases_ids = ",".join([str(tc.id) for tc in testcases])
-                testcases_order = TestcaseOrder.objects.create(
+                testcases_order = TestCaseOrder.objects.create(
                                     answer_paper=ans_paper,
                                     question=question,
-                                    testcase_order=testcases_ids)
+                                    order=testcases_ids)
             ans_paper.questions_order = ",".join(question_ids)
             ans_paper.save()
             ans_paper.questions_unanswered.add(*questions)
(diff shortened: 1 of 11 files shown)


maheshgudi:
@prathamesh920  Resolved all comments.

yaksh/templates/yaksh/view_answerpaper.html
```diff
@@ -34,7 +34,7 @@ <h2> Quiz: {{ paper.question_paper.quiz.description }} </h2>
     Start time: {{ paper.start_time }} <br/>
     End time : {{ paper.end_time }} <br/>
     Percentage obtained: {{ paper.percent }}% <br/>
-    {% if paper.passed == 0 %}
+    {% if paper.passed %}
     Status : <b style="color: red;"> Failed </b><br/>
     {% else %}
     Status : <b style="color: green;"> Passed </b><br/>
```
adityacp: This should be after if condition

yaksh/models.py
```diff
@@ -913,6 +919,17 @@ def get_test_case(self, **kwargs):
 
         return test_case
 
+    def get_ordered_test_cases(self, answerpaper):
+        try:
+            order = TestCaseOrder.objects.get(answer_paper=answerpaper,
+                                              question = self
+                                              ).order.split(",")
+            return [self.get_test_cases(id=int(tc_id))[0]\
```
adityacp: I guess there is no need of `\`

yaksh/templates/yaksh/view_answerpaper.html
maheshgudi: I have fixed this in PR #434

maheshgudi:
@adityacp resolved all comments.

yaksh/models.py
```diff
@@ -1182,7 +1199,17 @@ def make_answerpaper(self, user, ip, attempt_num, course_id):
...
             questions = self._get_questions_for_answerpaper()
             ans_paper.questions.add(*questions)
-            question_ids = [str(que.id) for que in questions]
+            question_ids = []
+            for question in questions:
+                question_ids.append(str(question.id))
+                testcases = question.get_test_cases()
+                if question.shuffle_testcases:
+                    random.shuffle(testcases)
+                testcases_ids = ",".join([str(tc.id) for tc in testcases])
```
prabhuramachandran: It is not clear why we need the order for every question even though shuffling makes no sense for assert, stdio, fill in the blanks etc.

prabhuramachandran:
Thanks but I would much prefer the shuffle MCQ option to be at the Question paper level and not the individual questions.  I.e. I don't want to set the individual questions to be shuffled or not but have the QP auto-shuffle all MCQ/MCCs if I decide to shuffle.

maheshgudi:
@prabhuramachandran Resolved all comments.

yaksh/models.py
```diff
@@ -1197,7 +1215,18 @@ def make_answerpaper(self, user, ip, attempt_num, course_id):
...
-            question_ids = [str(que.id) for que in questions]
+            question_ids = []
+            for question in questions:
+                question_ids.append(str(question.id))
+                testcases = question.get_test_cases()
+                if self.shuffle_testcases and \
+                    question.type in ["mcq", "mcc"]:
+                    random.shuffle(testcases)
+                testcases_ids = ",".join([str(tc.id) for tc in testcases])
+                testcases_order = TestCaseOrder.objects.create(
```
prathamesh920: I think what Sir pointed out here is why TestCaseOrder object is created for every questions! Since order is only needed for MCQ and MCC, so in TestCaseOrder, the order should be saved only for these question types.

maheshgudi:
@prathamesh920 resolved all comments.

PR #430, opened 2018-01-29
Fix multiple Answerpaper objects returned error

maheshgudi:
- Adds **user, question_paper, course and attempt number** as the **composite key** for answerpaper table.
- Adds test for the same

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -1161,7 +1161,7 @@ class QuestionPaper(models.Model):
         return all_questions
 
     def make_answerpaper(self, user, ip, attempt_num, course_id):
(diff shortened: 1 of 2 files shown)


yaksh/models.py
```diff
@@ -1503,6 +1495,9 @@ class AnswerPaper(models.Model):
 
     objects = AnswerPaperManager()
 
+    class Meta:
+        unique_together = ('user', 'question_paper', 'attempt_number')
```
adityacp: If the same quiz is added to two different courses then this will fail, so I think you need to add course also.

maheshgudi:
@adityacp Resolved all comments.

Issue #437, opened 2018-02-08
Grades App


prathamesh920:
- Need to add README
- Good UI
- There are fields like can_be_used and order in models for future use.
- More tests

Will separate out the issues as they are completed.

PR #438, opened 2018-02-08
Grades

### LINKED ISSUE #210: Grades and weightage for course
1. Moderator should be allowed to create a grading scheme for a particular course.
2. Each quiz within a course should have a quiz weightage.

prathamesh920:
Add weightage for Quiz and Create Grading System App

App Name: grades

Grading System provides with the grade for a given value.
It contains different grade ranges.

Has its own default grading system.
Allows you to modify and add grading system wth grade ranges.

To be done:
- Need to add README
- Good UI
- There are fields like can_be_used and order in models for future use.
- More tests
Created issue for the same #437

App name: Yaksh

Now every quiz has a default weightage of 100%, can be changed.
An aggregate is calculated for a given course.
Using grades app a grade is provide to the aggregate value.

Closes #210


grades/models.py
```diff
@@ -0,0 +1,49 @@
...
+        return self.name.title()
+
+
+class GradeRange(models.Model):
+    system = models.ForeignKey(GradingSystem)
+    order = models.IntegerField(default=0)
+    lower_limit = models.FloatField()
+    upper_limit = models.FloatField()
+    grade = models.CharField(max_length=10)
+    description = models.CharField(max_length=127, null=True, blank=True)
```
maheshgudi: Again, what does this do?

grades/views.py
```diff
@@ -0,0 +1,44 @@
...
+from django.contrib.auth.decorators import login_required
+from django.forms import inlineformset_factory
+from grades.forms import GradingSystemForm
+from grades.models import GradingSystem, GradeRange
+
+# Create your views here.
+@login_required
+def grading_systems(request):
+    user = request.user
+    default_grading_system = GradingSystem.objects.get(name='default')
```
maheshgudi: Problem is, all users share the same default model. If one user decides to play around with it, it will affect all users.

grades/views.py
```diff
@@ -0,0 +1,44 @@
...
+    grading_systems = GradingSystem.objects.filter(creator=user).exclude(
```
maheshgudi: exclude is not needed, since default is not created by user

yaksh/models.py
```diff
@@ -722,6 +725,33 @@ class CourseStatus(models.Model):
...
+            quizzes = self.course.get_quizzes()
+            total_weightage = 0
+            sum = 0
+            for quiz in quizzes:
+                total_weightage += quiz.weightage
+                marks = AnswerPaper.objects.get_user_best_of_attempts_marks(
+                        quiz, self.user.id, self.course.id)
+                out_of = quiz.questionpaper_set.first().total_marks
+                sum += (marks/out_of)*quiz.weightage
+            self.total_marks = (sum/total_weightage)*100
```
maheshgudi: This variable should be `percentage` instead of `total_marks`.

grades/models.py
```diff
@@ -0,0 +1,49 @@
+from django.db import models
+from django.contrib.auth.models import User
+
+# Create your models here.
+
+class GradingSystem(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(default='About the grading system!')
```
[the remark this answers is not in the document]
prathamesh920: Kept description because if an user wants to give some details regarding grading.

grades/models.py
```diff
@@ -0,0 +1,49 @@
...
+    can_be_used = models.BooleanField(default=False)
```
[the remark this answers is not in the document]
prathamesh920: Actually in PR description I have mentioned that fields ```can_be_used``` and ```order``` in models are for future use.
I had planned few things with these fields. These two are not used anywhere currently.
Will currently remove it in my next commit to avoid confusion.

grades/models.py
prathamesh920: Sometimes grade can have a short description like Outstanding, Excellent, Fail!

grades/views.py
prathamesh920: We are providing default grading system just for a reference, the only thing then we can do is not give permission to edit the default system.

grades/views.py
prathamesh920: Needed here because on editing default, while saving the default creator is that user!

grades/templates/grading_systems.html
```diff
@@ -0,0 +1,17 @@
+<html>
+    <b>Default Grading System:</b>
```
adityacp: I feel you should add a button or link to go back to moderator dashboard. It is difficult to navigate back to moderator dashboard

yaksh/templates/yaksh/course_modules.html
```diff
@@ -17,6 +17,7 @@
         <center>{{ msg }}</center>
     </div>
 {% endif %}
+<b>Grade: {% if grade %} {{ grade }} {% else %} Will be available once the course is complete {% endif %}</b>
```
adityacp: `Will be available once the course is complete` is not the proper message. since my course is completed but just because my obtained grade does not fit into available grading system this message shows up.

yaksh/models.py
```diff
@@ -720,8 +723,43 @@ class CourseStatus(models.Model):
...
+                out_of = quiz.questionpaper_set.first().total_marks
+                sum += (marks/out_of)*quiz.weightage
+            self.percentage = (sum/total_weightage)*100
+
+
+    def is_course_complete(self):
+        modules = self.course.get_learning_modules()
+        complete = False
+        for module in modules:
+            complete = module.get_status(self.user, self.course) == 'completed'
```
adityacp: Here I guess you have to check the condition if module's check_prerequisite is set to `True` otherwise it will return `False` even though all the other modules are completed.

yaksh/models.py
```diff
@@ -720,8 +723,43 @@ class CourseStatus(models.Model):
...
+            total_weightage = 0
+            sum = 0
+            for quiz in quizzes:
+                total_weightage += quiz.weightage
+                marks = AnswerPaper.objects.get_user_best_of_attempts_marks(
+                        quiz, self.user.id, self.course.id)
+                out_of = quiz.questionpaper_set.first().total_marks
+                sum += (marks/out_of)*quiz.weightage
+            self.percentage = (sum/total_weightage)*100
+
```
adityacp: I guess you have not added tests for `calculate_percentage` and `set_grade`

grades/templates/grading_systems.html
prathamesh920: I understand! But I have not kept Yaksh moderator dashboard link as its a separate grades app.
I m working on a way that it picks the base template of Yaksh by default.
It is include in issue #437 related to good UI

yaksh/templates/yaksh/course_modules.html
prathamesh920: For now it will pick the default grading system!
So grade will always be seen.
Need to add a way by which we can deactivate the grading system, will do with the above mentioned issue.

yaksh/models.py
prathamesh920: Yes because CourseStatus tests are not present in model tests and I am adding them but few problems I am facing currently so have mention in the issue that tests will be written.
Though will write tests for the methods.

yaksh/models.py
prathamesh920: I did not understand clearly what this means!
Please explain in detail.
I am simply using the method ```get_status``` to check the status!
Where and how check_prerequisite comes into picture here?

adityacp: There should be a way by which we can say not check a particular module's completion status similar to check_prerequisite because some modules might just be introduction modules containing no units. In your case, that module's completion status is also checked and since it is not completed the course will show `in progress`.  So, for now, `check_prerequisite` serves this purpose. So I thought we can add a condition to check if module `check_prerequisite` is set to True. We can discuss it.

prathamesh920: I think then it should be in module ```get_status``` method.
What do you think?
Because this is simply using the get_status method to check if the module is completed or not.
```get_status``` should return  completed if module has no unit!
May be we can add this in a different PR!
Created issue #440

adityacp: I have commented on #440 just check if that is ok to implement

grades/templates/grading_systems.html
adityacp: For now you can put a link with href having URL_ROOT/exam so that it redirects to dashboard later on you can change UI and make it better. Currently I have to hit back from browser back button.

prathamesh920: I am working in this with better implementation, will send with next PR. Not a major problem with the moderator.

prathamesh920:
Resolved all comments

prathamesh920:
@adityacp is creating the necessary UI.

PR #446, opened 2018-02-20
Fix clone course

### LINKED ISSUE #445: Change in clone course
The following change should be implemented for course clone:-
     - Shallow Copy Course (Create copy of modules and units)

adityacp:
This PR resolves issue mentioned in #445 
Now `Clone course` will clone all the modules and units along with the course.
Closes #445


yaksh/views.py
```diff
@@ -390,26 +390,27 @@ def prof_manage(request, msg=None):
...
         for answerpaper_id in delete_paper:
-            answerpaper = AnswerPaper.objects.get(id=answerpaper_id)
-            qpaper = answerpaper.question_paper
-            answerpaper.course.remove_trial_modules()
-            answerpaper.course.delete()
-            if qpaper.quiz.is_trial:
-                qpaper.quiz.delete()
-            else:
-                if qpaper.answerpaper_set.count() == 1:
+            answerpaper = AnswerPaper.objects.filter(id=answerpaper_id)
```
maheshgudi: Searching from id will always give only one object, why filter here?

yaksh/views.py
```diff
@@ -2245,17 +2246,26 @@ def download_sample_csv(request):
...
-def duplicate_course(request, course_id):
+def duplicate_course(request, copy_type, course_id):
     user = request.user
     course = Course.objects.get(id=course_id)
     if not is_moderator(user):
         raise Http404('You are not allowed to view this page!')
 
     if course.is_teacher(user) or course.is_creator(user):
-        course.create_duplicate_course(user)
+        if copy_type == "copy":
```
maheshgudi: should this be deep and shallow? ` copy` keyword as a copy_type is pretty ambiguous.

yaksh/views.py
prathamesh920: I think he wants to avoid ```try...except``` in case the object does not exists.

yaksh/templates/yaksh/courses.html
```diff
@@ -123,7 +133,7 @@
             </li>
             <br>
             <li>
-                <a href="{{URL_ROOT}}/exam/manage/duplicate_course/{{ course.id }}/">
+                <a class="test" href="{{URL_ROOT}}/exam/manage/duplicate_course/{{ course.id }}/" data-toggle="tooltip" title="Creates Copy of selected Course as well as its Modules, Lessons/Quizzes" data-placement="top">
```
maheshgudi: I am confused by the nomenclature of  the cloned course. Both deep and shallow cloning of courses are called Clone course. Should it be named differently? Also once a cloned copy is created, the course name is called Copy of *course*, irrespective of how it was cloned.

PR #452, opened 2018-03-01
Add a management command to add users to moderator group

### LINKED ISSUE #451: Management command to add users to Moderator group

ankitjavalkar:
- Adds a management command to add users to moderator group
- Rename ```add_group command``` to ```create_mod_group```

Closes #451


yaksh/management/commands/add_moderator.py
```diff
@@ -0,0 +1,33 @@
...
+
+class Command(BaseCommand):
+    help = 'Adds users to the moderator group'
+
+    def add_arguments(self, parser):
+        # Positional arguments
+        parser.add_argument('usernames', nargs='+', type=str)
+
+    def handle(self, *args, **options):
+        group = Group.objects.get(name='moderator')
```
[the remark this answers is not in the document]
ankitjavalkar: Agreed and fixed.

yaksh/management/commands/add_moderator.py
```diff
@@ -0,0 +1,33 @@
...
+        for uname in options['usernames']:
+            try:
+                user = User.objects.get(username=uname)
+                profile = user.profile
+            except User.DoesNotExist:
+                raise CommandError('User "{0}" does not exist'.format(uname))
+            except Profile.DoesNotExist:
+                raise CommandError('User "{0}" does not have a Profile'.format(uname))
+            group.user_set.add(user)
```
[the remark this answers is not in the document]
ankitjavalkar: Yes this is required. Added the necessary if condition.

ankitjavalkar:
Comments have been addressed.

yaksh/management/commands/add_moderator.py
```diff
@@ -0,0 +1,37 @@
...
+    help = 'Adds users to the moderator group'
+
+    def add_arguments(self, parser):
+        # Positional arguments
+        parser.add_argument('usernames', nargs='+', type=str)
+
+    def handle(self, *args, **options):
+        try:
+            group = Group.objects.get(name='moderator')
+        except Group.DoesNotExist:
```
maheshgudi: Why can't the group be created here? Is it a separate command? Why not make it a part of this?

yaksh/management/commands/add_moderator.py
```diff
@@ -0,0 +1,37 @@
...
+            raise CommandError('Moderator Group does not exist')
+
+        for uname in options['usernames']:
+            try:
+                user = User.objects.get(username=uname)
+            except User.DoesNotExist:
+                raise CommandError('User "{0}" does not exist'.format(uname))
+            if user in group.user_set.all():
+                self.stdout.write('User "{0}" is already a Moderator'.format(uname))
+                continue
```
maheshgudi: Why continue? Why unnecessarily add user to the group yet again? Couldn't the adding part be in an `else` condition?

yaksh/management/commands/add_moderator.py
ankitjavalkar: There is a separate command, it is preferable that both actions be kept individual.

yaksh/management/commands/add_moderator.py
ankitjavalkar: the ```continue``` command breaks out of that iteration of the loop and moves to the next iteration so that user is not added to the group and hence ```else``` is not required.

yaksh/management/commands/add_moderator.py
maheshgudi: Why? I don't know why it is preferable to have to separate commands. Could you explain it?

ankitjavalkar: In the current situation:
- There is already a command that adds moderators, so it is not required to add the same functionality here.
- If you refer to combining the two commands and dropping the ```create_mod_group``` command then that fails in the current scenario because the users are mandatory positional arguments, if the deployer tries to add moderator group without having any users in the database this command will fail.

ankitjavalkar: I will change the argument syntax so that the users can be added optionally and a moderator group is added by default.

maheshgudi: I second that idea. Will this be added in this PR itself?

PR #457, opened 2018-03-16
Specify tornado version to 4.5.3

maheshgudi:
1. The newer version of Tornado i.e v5.0.0 causes the current suite of code server tests to fail. Reverting the tornado version to 4.5.3.


prathamesh920:
Merging on priority basis.

PR #460, opened 2018-03-20
Display appropriate error when time between attempts condition is not satisfied

### LINKED ISSUE #459: Display appropriate error when time between attempts is not satisfied

ankitjavalkar:
Closes #459 

yaksh_attempt_error

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -317,7 +317,7 @@ class Quiz(models.Model):
     attempts_allowed = models.IntegerField(default=1, choices=attempts)
 
     time_between_attempts = models.FloatField(
-        "Time Between Quiz Attempts in hours"
+        "Time Between Quiz Attempts in hours", default=0.0
     )
 
     is_trial = models.BooleanField(default=False)
@@ -1356,11 +1356,18 @@ class QuestionPaper(models.Model):
             )
             if last_attempt:
                 time_lag = (timezone.now() - last_attempt.start_time).total_seconds() / 3600
-                return time_lag >= self.quiz.time_between_attempts
+                can_attempt = time_lag >= self.quiz.time_between_attempts
+                msg = "You cannot start the next attempt for this quiz before {0} hour(s)".format(
+                    self.quiz.time_between_attempts
+                ) if not can_attempt else None
+                return can_attempt, msg
             else:
-                return True
+                return True, None
         else:
-            return False
+            msg = "You cannot attempt {0} quiz more than {1} time(s)".format(
+                self.quiz.description, self.quiz.attempts_allowed
+            )
+            return False, msg
 
     def create_demo_quiz_ppr(self, demo_quiz, user):
         question_paper = QuestionPaper.objects.create(quiz=demo_quiz,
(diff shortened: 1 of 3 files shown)


yaksh/models.py
```diff
@@ -1216,11 +1216,18 @@ def can_attempt_now(self, user, course_id):
             )
             if last_attempt:
                 time_lag = (timezone.now() - last_attempt.start_time).total_seconds() / 3600
-                return time_lag >= self.quiz.time_between_attempts
+                can_attempt = time_lag >= self.quiz.time_between_attempts
```
maheshgudi: Instead this thing can be checked in views. This way this code remains intact.

yaksh/models.py
```diff
@@ -1216,11 +1216,18 @@ def can_attempt_now(self, user, course_id):
...
+                msg = "You cannot start the next attempt for this quiz before {0} hour(s)".format(
+                    self.quiz.time_between_attempts
+                ) if not can_attempt else None
+                return can_attempt, msg
             else:
-                return True
+                return True, None
```
maheshgudi: This is why doing it in views make sense. Sending None is pointless

ankitjavalkar: I don't think that it is pointless, ```None``` just means that there is no error message. Pushing this to the views will mean adding if statements in a place that is already crowded with a lot of decision making in the first place (lines 543 to line 559 or so).

Returning all the necessary data from the models keeps the views clean.

yaksh/models.py
ankitjavalkar: I just changed the return values, earlier it used to be just a True/False now the error message also gets set here.

yaksh/views.py
```diff
@@ -532,9 +532,10 @@ def start(request, questionpaper_id=None, attempt_num=None, course_id=None,
             previous_question=last_attempt.current_question()
         )
     # allowed to start
-    if not quest_paper.can_attempt_now(user, course_id):
-        msg = "You cannot attempt {0} quiz more than {1} time(s)".format(
-            quest_paper.quiz.description, quest_paper.quiz.attempts_allowed)
+    if not quest_paper.can_attempt_now(user, course_id)[0]:
+        # msg = "You cannot attempt {0} quiz more than {1} time(s)".format(
```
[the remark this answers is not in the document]
ankitjavalkar: Will remove this along with addition of test cases

yaksh/models.py
maheshgudi: Fair enough.

maheshgudi: I just believe that a method shouldn't return a message. To me, I don't believe model methods should do that.

yaksh/views.py
```diff
@@ -534,9 +534,8 @@ def start(request, questionpaper_id=None, attempt_num=None, course_id=None,
             previous_question=last_attempt.current_question()
         )
     # allowed to start
-    if not quest_paper.can_attempt_now(user, course_id):
-        msg = "You cannot attempt {0} quiz more than {1} time(s)".format(
-            quest_paper.quiz.description, quest_paper.quiz.attempts_allowed)
+    if not quest_paper.can_attempt_now(user, course_id)[0]:
+        msg = quest_paper.can_attempt_now(user, course_id)[1]
```
maheshgudi: We can add the message here. There is not a whole lot of reason for the msg to be generated inside the method

PR #461, opened 2018-03-21
Quiz Completion Status

### LINKED ISSUE #464: Quiz progress issue
Quiz status shows in progress even if the quiz is completed.

adityacp:
This PR fixes an issue where quiz completion status is shown as not completed even if it is attempted.
Following two conditions will be checked for quiz completion:-
1. If a quiz is present in completed units
2. If a quiz has an answer paper then get the answer paper status i.e `completed` or `in progress`. If there is no answer paper then the quiz is considered as `not attempted`.

Closes #464


yaksh/models.py
```diff
@@ -383,6 +383,20 @@ def get_failed_students(self, course):
                 course=course, passed=False
             ).values_list("user", flat=True).distinct().count()
 
+    def get_answerpaper_status(self, user, course):
+        try:
+            qp = self.questionpaper_set.get().id
+        except QuestionPaper.DoesNotExist:
+            qp = None
```
prathamesh920: Here itself if I ```return "not attempted" ``` will be better!

yaksh/models.py
```diff
@@ -383,6 +383,20 @@ def get_failed_students(self, course):
                 course=course, passed=False
             ).values_list("user", flat=True).distinct().count()
 
+    def get_answerpaper_status(self, user, course):
```
prathamesh920: ```AnswerPaper``` model already has ```status``` attribute, so the name ```get_answerpaper_status``` is not proper here.

yaksh/models.py
```diff
@@ -383,6 +383,20 @@ def get_failed_students(self, course):
...
+        try:
+            qp = self.questionpaper_set.get().id
+        except QuestionPaper.DoesNotExist:
+            qp = None
+        ans_ppr = AnswerPaper.objects.filter(
+            user=user, course=course, question_paper=qp
+        )
+        if ans_ppr.exists():
+            status = "completed"
```
prathamesh920: Can state be ```inprogress```?

yaksh/models.py
```diff
@@ -421,6 +435,8 @@ def get_completion_status(self, user, course):
         if course_status.exists():
             if self in course_status.first().completed_units.all():
                 state = "completed"
+            elif self.type == "quiz":
```
prathamesh920: So the issue is when quiz is finished and the user does not click the next learning unit, the learning unit is not considered as completed, but the Quiz is complete.
Am I correct about the issue?

If yes then, try if possible to update learning unit along with the completion of Quiz!

PS: Please refer issue number in the PR

adityacp: Can you suggest where can I add a quiz to completion unit?

yaksh/models.py
maheshgudi: Can it boolean? String comparison isn't the cleanest of ways to do this!

yaksh/models.py
adityacp: Resolved. Now quiz completion will depend on the answer paper status `in progress` or `completed`.

yaksh/models.py
adityacp: Now `get _answerpaper_status` will return answerpaper status i.e in progress or completed.

yaksh/test_models.py
```diff
@@ -708,6 +709,9 @@ def setUpClass(self):
         self.trial_course = Course.objects.create_trial_course(self.user)
         self.trial_quiz = Quiz.objects.create_trial_quiz(self.user)
 
+    @classmethod
+    def tearDownClass(self):
+        self.quiz.questionpaper_set.all().delete()
```
maheshgudi: Deleting trial course should delete everything associated with it. I believe deleting trial course is cleaner. Not really a problem though

yaksh/models.py
```diff
@@ -383,6 +383,20 @@ def get_failed_students(self, course):
...
+        try:
+            qp = self.questionpaper_set.get().id
+        except QuestionPaper.DoesNotExist:
+            qp = None
+        ans_ppr = AnswerPaper.objects.filter(
+            user=user, course=course, question_paper=qp
+        ).order_by("-attempt_number")
+        if ans_ppr.exists():
+            status = ans_ppr.first().status
+        else:
```
maheshgudi: Can't we just let status be returned as `None`? we can check that instead of checking `if status == "not attempted"?`

adityacp: This `not attempted` is used to show on the interface. So I feel it is okay to send from models. If I make it None then I need to add a condition in the template.

adityacp: As explained previously, `not attempted` is used at various places to show the status of a quiz or lesson or module. So instead of adding conditions in multiple templates, I directly pass the status from model methods.

yaksh/test_models.py
maheshgudi: Ok. Nevermind then!

PR #466, opened 2018-03-23
Remove the context instance argument in views.py

ankitjavalkar:
- Remove the context instance argument in views.py (deprecated since Django 1.10 https://docs.djangoproject.com/en/1.10/releases/1.10/)
- Replace ```render_to_response``` with ```render```

### DIFF
--- a/yaksh/decorators.py
+++ b/yaksh/decorators.py
@@ -1,4 +1,4 @@
-from django.shortcuts import render_to_response, redirect
+from django.shortcuts import render, render_to_response, redirect
 from django.conf import settings
 from django.template import RequestContext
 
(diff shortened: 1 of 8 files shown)


yaksh/decorators.py
```diff
@@ -1,4 +1,4 @@
-from django.shortcuts import render_to_response, redirect
+from django.shortcuts import render, render_to_response, redirect
```
prathamesh920: ```render_to_response``` is replaced by ```render```, then ```render_to_response``` can be removed from the import as it is not used any more.

yaksh/views.py
```diff
@@ -7,7 +7,7 @@
 from django.http import HttpResponse, JsonResponse
 from django.core.urlresolvers import reverse
 from django.contrib.auth import login, logout, authenticate
-from django.shortcuts import render_to_response, get_object_or_404, redirect
+from django.shortcuts import render_to_response, render, get_object_or_404, redirect
```
prathamesh920: Again ```render_to_response``` needs to be removed.

ankitjavalkar:
Fixed changes as per suggestions

ankitjavalkar:
I have bumped the django version from 1.9.5 to 1.10. The Context instance is deprecated since v1.10, keeps giving warnings on test-server/local servers.

PR #469, opened 2018-04-04
Enhance course status

### LINKED ISSUE #468: Course Progress loads slowly
Currently course progress page loads slowly for a large number of users.

adityacp:
Closes #468 

Course Status Per User for moderator and teacher
student_course_status

Student Dashboard with course and module progress
student_dashboard

### DIFF
--- a/yaksh/views.py
+++ b/yaksh/views.py
@@ -488,9 +488,6 @@ def start(request, questionpaper_id=None, attempt_num=None, course_id=None,
                 learning_module.name)
             return course_modules(request, course_id, msg)
 
-    # update course status with current unit
-    _update_unit_status(course_id, user, learning_unit)
-
     # is user enrolled in the course
     if not course.is_enrolled(user):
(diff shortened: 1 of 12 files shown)


yaksh/views.py
```diff
@@ -2759,29 +2758,36 @@ def course_status(request, course_id):
...
-            user=user, course_id=course_id
-        )
-    else:
-        course_status = course_status.first()
     # make next available unit as current unit
     course_status.current_unit = unit
     course_status.save()
 
 
+def _update_course_percent(course, user):
```
prathamesh920: Can't this be a model method?

adityacp: If I make this a model method then anyhow I have to write the Course status query every time. Instead of that, I created this view function so that this can be reused anywhere. I don't need to write course status query every time. All the other updates are being done by model methods.

adityacp:
Resolved all the comments

PR #471, opened 2018-04-05
Python Assertion Evaluation additions and changes

### LINKED ISSUE #470: Python Assertion Evaluation additions and changes
1. Along With all the error, the output should show whatever test case is tried on student code.
2. Add nose tools in global space so that nose asserts can be used in the standard test case.

adityacp:
Closes #470

### DIFF
--- a/yaksh/python_assertion_evaluator.py
+++ b/yaksh/python_assertion_evaluator.py
@@ -39,7 +39,6 @@ class PythonAssertionEvaluator(BaseEvaluator):
             submitted = compile(self.user_answer, '<string>', mode='exec')
             self.exec_scope = {}
             exec(submitted, self.exec_scope)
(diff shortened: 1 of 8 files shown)


yaksh/python_assertion_evaluator.py
```diff
@@ -43,6 +39,7 @@ def compile_code(self):
             submitted = compile(self.user_answer, '<string>', mode='exec')
             self.exec_scope = {}
             exec(submitted, self.exec_scope)
+            exec("from nose.tools import *", self.exec_scope)
```
ankitjavalkar: Please don't use star, since we do not know what this scope will include given that student code could be anything.

adityacp: Nose is added to the scope for the moderator so that he can use nose asserts on student's code

PR #480, opened 2018-05-16
[Role based implementation] Allow moderator to switch between student and moderator roles

### LINKED ISSUE #478: Allow Moderator to switch to Student Role and back

ankitjavalkar:
Closes #478

### DIFF
--- a/yaksh/evaluator_tests/test_simple_question_types.py
+++ b/yaksh/evaluator_tests/test_simple_question_types.py
@@ -1,6 +1,7 @@
 import unittest
 from datetime import datetime, timedelta
 from django.utils import timezone
+from django.contrib.auth.models import Group
 from textwrap import dedent
 import pytz
 from yaksh.models import User, Profile, Question, Quiz, QuestionPaper,\
@@ -8,7 +9,9 @@ from yaksh.models import User, Profile, Question, Quiz, QuestionPaper,\
     StringTestCase, McqTestCase, ArrangeTestCase
 
 
+
 def setUpModule():
+    mod_group = Group.objects.create(name='moderator')
     # Create user profile
     # Create User 1
     user = User.objects.create_user(username='demo_user_100',
@@ -46,6 +49,7 @@ def setUpModule():
 def tearDownModule():
     User.objects.filter(username__in=["demo_user_100", "demo_user_101"])\
                  .delete()
+    Group.objects.all().delete()
 
 
 class IntegerQuestionTestCases(unittest.TestCase):
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -8,7 +8,7 @@ from ruamel.yaml.comments import CommentedMap
 from random import sample
 from collections import Counter
 from django.db import models
-from django.contrib.auth.models import User
+from django.contrib.auth.models import User, Group
 from django.contrib.contenttypes.models import ContentType
 from taggit.managers import TaggableManager
 from django.utils import timezone
@@ -982,6 +982,7 @@ class Profile(models.Model):
     institute = models.CharField(max_length=128)
     department = models.CharField(max_length=64)
     position = models.CharField(max_length=64)
+    is_moderator = models.BooleanField(default=False)
     timezone = models.CharField(
         max_length=64,
         default=pytz.utc.zone,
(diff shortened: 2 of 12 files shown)


yaksh/views.py
```diff
@@ -1730,6 +1739,30 @@ def search_teacher(request, course_id):
...
+
+    try:
+        group = Group.objects.get(name='moderator')
+    except Group.DoesNotExist:
+        raise Http404('The Moderator group does not exist')
+
+    if not user.profile.is_moderator:
+        raise Http404('You are not allowed to view this page!')
+
+    print group.user_set.all()
```
adityacp: why print?

ankitjavalkar: Removed

yaksh/test_views.py
```diff
@@ -5063,6 +5116,16 @@ def setUp(self):
...
+        Profile.objects.create(
+            user=self.teacher,
+            roll_number=10,
+            institute='IIT',
+            department='Chemical',
+            position='Moderator',
+            timezone='UTC',
+            is_moderator=True
+        )
+
```
prathamesh920: It will be better if the tests are added for the new view created.

yaksh/views.py
```diff
@@ -75,18 +75,27 @@ def my_render_to_response(request, template, context=None, **kwargs):
     return render(request, template, context, **kwargs)
 
 
-def is_moderator(user):
+def is_moderator(user,  group_name='moderator'):
```
prathamesh920: The keyword argument ```group_name``` is not used in the function.

yaksh/views.py
```diff
@@ -1730,6 +1739,29 @@ def search_teacher(request, course_id):
...
+    except Group.DoesNotExist:
+        raise Http404('The Moderator group does not exist')
+
+    if not user.profile.is_moderator:
+        raise Http404('You are not allowed to view this page!')
+
+    if user not in group.user_set.all():
+        group.user_set.add(user)
+    else:
+        group.user_set.remove(user)
```
prathamesh920: I think the moderator is frequently added or removed from the group is not a good idea!

ankitjavalkar:
Closing this PR and sending a new one with the suggested implementation.

ankitjavalkar:
Need to reach a decision whether to use implementation in #492  or the one in #480

yaksh/models.py
```diff
@@ -998,6 +1016,17 @@ def get_user_dir(self):
...
         return user_dir
 
+    def _toggle_moderator_group(self, group_name):
+        group = Group.objects.get(name=group_name)
+        if self.is_moderator:
+            self.user.groups.add(group)
+        else:
+            self.user.groups.remove(group)
+
+    def save(self, *args, **kwargs):
```
prathamesh920: ```save``` is overridden here!
And toggle is called for every save!

In case the moderator switches he is no longer in the group, now he is a student, first he edits his profile then save is called with toogle and the student will become moderator again!

yaksh/templates/yaksh/quizzes_user.html
```diff
@@ -7,6 +7,13 @@
             <center>{{ msg }}</center>
         </div>
     {% endif %}
+    {% if user.profile.is_moderator %}
+        <div class="row">
+            <a href="{{URL_ROOT}}/exam/toggle_moderator/" class="btn btn-primary pull-right" style="margin-top: 20px; margin-bottom: 20px;">
+              Switch To Moderator
```
prathamesh920: I feel the buttons to switch should be on the top nav bar. Always visible to the users.
If majority are fine with this then I am okay with it.

yaksh/evaluator_tests/test_simple_question_types.py
```diff
@@ -1,14 +1,17 @@
...
 from textwrap import dedent
 import pytz
 from yaksh.models import User, Profile, Question, Quiz, QuestionPaper,\
     AnswerPaper, Answer, Course, IntegerTestCase, FloatTestCase,\
     StringTestCase, McqTestCase, ArrangeTestCase
 
 
+
 def setUpModule():
+    mod_group = Group.objects.create(name='moderator')
```
prathamesh920: Does this need to be assigned?

ankitjavalkar: Yes, test cases fail otherwise.

yaksh/templates/yaksh/quizzes_user.html
ankitjavalkar: In moderator templates, the nav bar becomes very very crowded because the button has 3 words.

yaksh/models.py
ankitjavalkar: Good catch, fixed.

prathamesh920:
LGTM

PR #487, opened 2018-06-26
Download course content for offline view

### LINKED ISSUE #486: Course Download for offline view
Moderator and Student should be able to download the course content for offline view.

adityacp:
Closes #486


yaksh/templates/yaksh/courses.html
```diff
@@ -140,6 +140,11 @@
                 <a class="test" href="{{URL_ROOT}}/exam/manage/duplicate_course/{{ course.id }}/" data-toggle="tooltip" title="Creates Copy of selected Course as well as its Modules, Lessons/Quizzes" data-placement="top">
                 Clone Course</a>
             </li>
+            <br>
+            <li>
+                <a href="{{URL_ROOT}}/exam/manage/courses/download_course/{{course.id}}" data-toggle="tooltip" title="Download course content in html format" data-placement="top">Download Course
```
maheshgudi: Should be Download for Offline viewing or something like that! Just Download Course with Download Course csv and Download Yaml course will be pretty confusing!!

yaksh/views.py
```diff
@@ -2808,3 +2816,27 @@ def get_user_data(request, course_id, student_id):
         context = Context(data)
         data = template.render(context)
     return HttpResponse(json.dumps({"user_data": data}), **response_kwargs)
+
+
+@login_required
+@email_verified
+def download_course(request, course_id):
```
maheshgudi: Should be offline_view_course or download_offline_course

yaksh/templates/yaksh/unit.html
```diff
@@ -0,0 +1,82 @@
+{% load custom_filters %}
+<html>
+<link rel="stylesheet" href="../../static/css/bootstrap.css">
+<link rel="stylesheet" href="../../static/css/bootstrap.min.css">
+<link rel="stylesheet" href="../../static/css/video-js.css">
+<script src="../../static/js/jquery-1.9.1.min.js"></script>
+<script src="../../static/js/bootstrap.min.js"></script>
+<script src="../../static/js/video.js"></script>
+<style>
+.footer {
```
maheshgudi: This is bad!!! Please add this in a different css file.

yaksh/file_utils.py
```diff
@@ -66,3 +67,31 @@ def is_csv(document):
...
+        for file in static_files[folder]:
+            file_path = os.sep.join((folder_path, file))
+            with open(file_path, "rb") as f:
+                zipfile.writestr(
+                    os.sep.join((course_name, "static", folder, file)),
+                    f.read()
+                    )
+
+
+def write_templates_to_zip(zipfile, template_path, data, filename, filepath):
```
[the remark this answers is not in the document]
adityacp: Cannot add static files in this function because static files will be added to the zip file only once. Whereas this function is called multiple times to add the different templates to the zip file.

yaksh/file_utils.py
```diff
@@ -66,3 +67,31 @@ def is_csv(document):
...
+    with open(template_path) as f:
+        template_data = f.read()
+        template = Template(template_data)
```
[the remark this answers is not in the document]
adityacp: This is already generic I feel just need to make the zip_file argument as None. Then this can be used to render error messages.

yaksh/views.py
adityacp: I feel this is ok because the same function can be used to download course with YAML content with another argument as a type which can either be `html` or `yaml`. In that case suggested function name will not be appropriate.

yaksh/file_utils.py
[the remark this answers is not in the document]
maheshgudi: Can all templates be rendered first and then added with the static files?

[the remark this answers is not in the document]
prathamesh920: Just an idea that static argument default value is None.

yaksh/models.py
```diff
@@ -124,7 +127,10 @@ def dict_to_yaml(dictionary):
 
 
 def get_file_dir(instance, filename):
-    upload_dir = instance.lesson.name.replace(" ", "_")
+    if isinstance(instance, LessonFile):
```
prathamesh920: Just suggestion.
In future I feel will be better if we have a separate function, for each instances.
Eg: get_lesson_dir, get_video_dir ...
This will reduce the ifs in a single function.
Also if a new person is implementing similar thing then may write his own callable as per the need and no need to worry about the use of the existing function.

yaksh/urls.py
```diff
@@ -200,5 +200,9 @@
     url(r'^manage/preview_questionpaper/(?P<questionpaper_id>\d+)/$',
         views.preview_questionpaper, name="preview_questionpaper"),
     url(r'^manage/get_user_status/(?P<course_id>\d+)/(?P<student_id>\d+)/$',
-        views.get_user_data, name="get_user_data")
+        views.get_user_data, name="get_user_data"),
+    url(r'^manage/courses/download_course/(?P<course_id>\d+)/$',
+        views.download_course, name="download_course"),
+    url(r'^download_course/(?P<course_id>\d+)/$',
+        views.download_course, name="download_course"),
```
prathamesh920: Both urls are similar and they have same name!
Same name is fine if there is difference in the arguments.
Anyway here it wont harm as as nothing special is happening for the moderator.
And we try to use {% url %} tag or reserve match then in case of tie it will take the last url.
Its fine for now, in future we need to be careful.

yaksh/models.py
prathamesh920: No issue here.
Just that two different instances were seen in the same function, thought of mentioning.

prathamesh920:
Merging on priority basis

PR #492, opened 2018-07-05
Allow moderators to access student view

ankitjavalkar:


ankitjavalkar:
Need to reach a decision whether to use implementation in #492 or the one in #480

PR #496, opened 2018-07-12
Question upload can take yaml and zip files

### LINKED ISSUE #339: Question Upload expects a Zip file
When uploading questions, you always have to put it in a zip file and add a questions_dump.yaml which is irritating.  I should be able to upload any yaml file *or* a zip file.  It should pick up all yaml files in the dumped zip file and use those and not expect a particular naming.

maheshgudi:
- Moderators can now upload individual yaml file containing questions instead of zip files. Also the yaml file can have any name.
- Moderators can now upload zip file containing multiple yaml files. No specific name is required for any yaml or the zip file. Also, files used by certain questions have to be **specifically added inside a folder which is on the same level as yaml files**. These folders containing additional files can be of any name and number.
- Added information panel on upload questions for better understanding.
- Added test cases for the aforementioned enhancements.
- Update docs.
- closes #339


adityacp:
LGTM

PR #497, opened 2018-07-12
Change README

adityacp:
Change command for installing dependencies during local deployment.
Change production deployment instructions with a proper order to follow.

### DIFF
--- a/README_production.rst
+++ b/README_production.rst
@@ -78,8 +78,6 @@ To install this app follow the steps below:
 
    ::
 
-        $ python manage.py makemigrations
-
         $ python manage.py migrate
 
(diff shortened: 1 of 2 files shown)


README_production.rst
```diff
@@ -58,52 +50,67 @@ To install this app follow the steps below:
...
+      DB_PASSWORD=mypassword # Or the password used while creating a Database
+      DB_PORT=3306
 
-4. Run:
+5. Run:
 
    ::
 
-       $ python manage.py makemigrations yaksh
+        $ python manage.py makemigrations
```
prathamesh920: We don't need to do makemigrations

PR #502, opened 2018-07-17
Fix assignment upload

adityacp:
Currently, uploaded files for assignments are stored per quiz. This can cause issues if the same quiz is added to separate courses.
So now along with the quiz, the files are uploaded and stored per course.

### DIFF
--- a/yaksh/views.py
+++ b/yaksh/views.py
@@ -2098,14 +2098,14 @@ def update_email(request):
 
 @login_required
 @email_verified
-def download_assignment_file(request, quiz_id, course_id=None,
+def download_assignment_file(request, quiz_id, course_id,
                              question_id=None, user_id=None):
     user = request.user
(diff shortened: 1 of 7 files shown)


yaksh/views.py
```diff
@@ -2100,13 +2098,18 @@ def update_email(request):
 
 @login_required
 @email_verified
-def download_assignment_file(request, quiz_id, question_id=None, user_id=None):
+def download_assignment_file(request, quiz_id, course_id=None,
+                             question_id=None, user_id=None):
     user = request.user
-    if not is_moderator(user):
-        raise Http404("You are not allowed to view this page")
+    course = Course.objects.get(id=course_id)
```
maheshgudi: This will fail at some point, given that course_id can be none and get will fail!

adityacp:
You can merge the PR if it is OK. Migrations changes will be pushed in the release PR.

PR #503, opened 2018-07-18
Add a redirect from url / to /exam

ankitjavalkar:


ankitjavalkar:
Unfortunately, that's the most flexible way to do it. Doing it form the app itself means that this app will no longer remain pluggable. All requests to index URL '/' will then always be redirected to 'exam', In cases where I would want to use this as a part of a larger project that is not desired.

online_test/views.py
```diff
@@ -0,0 +1,7 @@
+from django.shortcuts import redirect
+
+from .settings import URL_ROOT
+
+
+def index(request):
+    return redirect(URL_ROOT+'/exam/')
```
adityacp: Instead of plus you can use format.

ankitjavalkar:
Merging on priority basis.

PR #507, opened 2018-07-24
Remove references to non existent static CSS files

### LINKED ISSUE #506: Multiple CSS files do not exist - Causes Internal Server Error
Multiple CSS files do not exist or have been deprecated and hence cause internal server errors, one such example:

```/yaksh/templates/yaksh/addteacher.html``` references a non existent CSS file which causes internal server errors.

```
{% block css %}
<link rel="stylesheet" media="all" type="text/css" href="{{ URL_ROOT }}/static/yaksh/css/course.css" />
{% endblock %}
```
Server Error:

```
request: "GET /static/yaksh/css/course.css HTTP/2.0", host: "*****.fossee.in", referrer: "https://*****.fossee.in/exam/manage/searchteacher/5/"
```

ankitjavalkar:
- Remove reference to /static/yaksh/css/course.css in addteacher.html
- Remove reference to /static/yaksh/css/question_quiz.css in add_course, add_exercise and add_quiz

Closes #506


ankitjavalkar:
I am removing ```showusers.html``` based on a quick cursory search that seems to indicate this is not being used anywhere, similarly the associated CSS file does not exist.

Cannot seem to catch any other redendants.

prathamesh920:
Also ```ajax_marks.html``` is not used anymore.

In ```question_paper_creation.js```  there is a comment saying "ajax request..." at line 6 but
it is not an ajax call, so it can be removed.

ankitjavalkar:
Done.

PR #518, opened 2018-07-28
Add -lm to the link flags for math functions

kumanna:
Some math functions like fabs are optimized by gcc, and these don't
require linking against libm.so. However, for most common functions
such as sqrt, we need to link against libm.so. This commit adds the -lm
flag to enable this.

### DIFF
--- a/yaksh/cpp_code_evaluator.py
+++ b/yaksh/cpp_code_evaluator.py
@@ -51,9 +51,9 @@ class CppCodeEvaluator(BaseEvaluator):
 
     def get_commands(self, clean_ref_code_path, user_output_path,
                      ref_output_path):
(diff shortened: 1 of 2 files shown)


ankitjavalkar:
Since you are reverting to g++, does that mean that now the student-submitted code will handle the situations caused by name mangling?
Just wanted to understand the reason behind your decision to revert to g++.

kumanna:
Reverted the g++ shift. Forgot completely about the mangling issues. I did that since the tests were failing, but then,  I guess the tests have to be rewritten to handle this. Thanks for catching this.

A better approach would be to handle C and C++ separately. But this is longer term and not important for me at this time.

ankitjavalkar:
Yes, we can ignore the failing tests for now. I'll merge and update the instance first thing on Monday.

PR #528, opened 2018-11-10
Fix #525

### LINKED ISSUE #525: Legend of Question Types in Quiz (UI)
Implementation of the legend in the running Quiz seems to be slightly unpolished in the new UI. May be this could be made better.

CruiseDevice:
- Categorize questions into objectives and programming.


adityacp:
Just a suggestion instead of using custom filters why can't you pass the question categories from views when the quiz is created using simple Django filter instead of looping through the lists.

yaksh/templatetags/custom_filters.py
```diff
@@ -91,7 +91,24 @@ def replace_spaces(name):
...
+def get_question_by_index(all_questions, questions_by_type):
+    return [index for index, item in enumerate(all_questions)
+            if item in set(questions_by_type)]
+
+
+@register.simple_tag
+def get_categories(question_types):
+    categories = {}
+    objectives = [type_ for type_ in question_types if type_ != 'code']
+    programming = [type_ for type_ in question_types if type_ == 'code']
```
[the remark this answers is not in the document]
CruiseDevice:
```
categories = {}
    categories["objectives"] = []
    categories["programming"] = []
    for type_ in all_question_types:
        if type_ != 'code':
            categories["objectives"].append(type_)
        else:
            categories["programming"].append(type_)
```
Using this in views.py instead of custom_filter.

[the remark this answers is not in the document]
adityacp: No need of this loop instead use the filter to get questions categories.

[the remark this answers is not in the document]
prathamesh920: I agree with @adityacp 
I also feel that the code is too complex, can be simplified.
Also, the assignment upload question is mainly non-objective type, so need to check that as well!
AnswerPaper stores the order of the questions. You can also think of having a class method that will simply give you all the objectives and the code questions for a given answerpaper.
So in the template you don't need to process anything complex or even the filter!

yaksh/static/yaksh/css/yakshcustom.css
```diff
@@ -229,4 +229,41 @@ html {
...
+}
+
+.div-table-header{
+	display: table-cell;
+	font-weight: bold;
+	padding:10px;
+	border-bottom: 1px solid;
+	border-left: 1px solid;
+	border-right: 1px solid;
+}
```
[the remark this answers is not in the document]
CruiseDevice: Yes table class css in `exam.css` is breaking the table. After commenting the css it is displaying the table properly. Also i don't think `exam.css` is being used anywhere other than in `question.html`. Also when i looked in `question.html` there was no table in it.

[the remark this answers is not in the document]
CruiseDevice: Yes table class css in `exam.css` is breaking the table. After commenting the css it is displaying the table properly. Also i don't think `exam.css` is being used anywhere other than in question.html. Also when i looked in `question.html` there was no table in it.

yaksh/templatetags/custom_filters.py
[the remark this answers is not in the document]
CruiseDevice: @adityacp i came up with something like this 
```
    questions = paper.question_paper.fixed_questions.all()
    objectives = questions.all().exclude(type='code').exclude(type='upload')
    upload = questions.filter(type='upload')
    programming = questions.filter(type='code')
```

yaksh/views.py
```diff
@@ -616,6 +631,44 @@ def show_question(request, question, paper, error_message=None,
     module = course.learning_module.get(id=module_id)
     all_modules = course.get_learning_modules()
     all_question_types = [types[0] for types in question_types]
+    categories = {}
+    objectives = questions.filter(Q(type="mcq") |
```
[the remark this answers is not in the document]
CruiseDevice: How about creating a dictionary for categories, then creating a separate dict for types and then using these two to get question index. I am also thinking of creating another dictionary for maintaining question index according to category. Just let me know whats on your mind.

yaksh/static/yaksh/css/exam.css
```diff
@@ -1,7 +1,13 @@
...
 			}
 #stdio, #assertion {
 	table-layout: fixed
 }
+.table{
+	border: 1px solid;
+	background-color: white;
+	border-spacing: 5px;
+	border-collapse: collapse;
+}
```
maheshgudi: Add a new line here. Although not a deal breaker, its better to develop a habit for this!

yaksh/static/yaksh/css/yakshcustom.css
```diff
@@ -227,6 +227,4 @@ html {
     max-height: 500px;
     overflow-y: auto;
     -ms-overflow-style: -ms-autohiding-scrollbar;
-}
-
-/*---------------------*/
\ No newline at end of file
+}
```
maheshgudi: Same here

yaksh/models.py
```diff
@@ -2086,6 +2086,21 @@ def is_attempt_inprogress(self):
     def get_previous_answers(self, question):
         return self.answers.filter(question=question).order_by('-id')
 
+    def get_question_by_index(self, all_ordered_questions, questions_by_type):
```
prathamesh920: This can be a private method.

prathamesh920: Also, ```all_ordered_questions``` need not be passed as an argument, since it's an answer paper model method you can get it for each instance using ```self.get_all_ordered_questions```

yaksh/models.py
```diff
@@ -2086,6 +2086,21 @@ def is_attempt_inprogress(self):
...
+        return [index+1 for index, item in enumerate(all_ordered_questions)
+                if item in set(questions_by_type)]
+
+    def get_questions_by_type(self, all_ordered_questions, objective_types):
```
prathamesh920: Why ```objective_types```?
It can be any type, right?

yaksh/static/yaksh/css/exam.css
```diff
@@ -1,7 +1,13 @@
-table td, table th { border: #ff8295 solid 1px !important;
+#assertion td, #assertion th, #stdio td, #stdio th { border: #ff8295 solid 1px !important;
 			word-wrap: break-word !important;
 			white-space: pre-wrap !important;
 			}
 #stdio, #assertion {
 	table-layout: fixed
 }
+.table{
```
prathamesh920: Give a different name if custom CSS is needed!
Because table class provided by bootstrap will overwritten if used by any one else in the same file in future.

yaksh/templates/exam.html
```diff
@@ -83,28 +78,56 @@
...
+              <tr>
+                <td>
+                  Programming:
+                </td>
+                <td>
+                  {{index.programming_index|join:", "}}
+                </td>
+              </tr>
+              {% endif %}
+            </tbody>
```
prathamesh920: I think it will be more than enough if we simply show objectives and programming!
Blanks are also objectives, and objective types will keep on adding in future, so better keep it simple for a student as well.
Upload type can be separate.

yaksh/views.py
```diff
@@ -616,6 +617,54 @@ def show_question(request, question, paper, error_message=None,
...
+    types["objective_types"] = set([_type.type
+                                    for _type in
+                                    categorized_questions["objectives"]])
+    types["blank_types"] = set([_type.type
+                                for _type in
+                                categorized_questions["blanks"]])
+    types["programming_types"] = set([_type.type
+                                      for _type in
+                                      categorized_questions["programming"]])
+    types["upload_types"] = set([_type.type
```
prathamesh920: No need of doing this everytime!
Simply create a constants that contains the objective types and programming types.
Also the above filters are then not needed.
This will be good as everytime a new type is added, then we don't need to modify this code, simply add to the constant!

yaksh/views.py
```diff
@@ -616,6 +617,54 @@ def show_question(request, question, paper, error_message=None,
...
+    categorized_questions["objectives"] = questions.filter(Q(type="mcq") |
+                                                           Q(type="mcc") |
+                                                           Q(type="arrange"))
+    categorized_questions["blanks"] = questions.filter(Q(type="integer") |
+                                                       Q(type="string") |
+                                                       Q(type="float"))
+    categorized_questions["programming"] = questions.filter(Q(type="code"))
+    categorized_questions["upload"] = questions.filter(Q(type="upload"))
+
+    types["objective_types"] = set([_type.type
```
prathamesh920: ```_type``` should be ```question```

ankitjavalkar:
Looks good, can be merged.

yaksh/views.py
```diff
@@ -568,6 +568,7 @@ def show_question(request, question, paper, error_message=None,
                   previous_question=None):
     """Show a question if possible."""
     quiz = paper.question_paper.quiz
+    questions = paper.questions.all()
```
prathamesh920: Why is this needed?

yaksh/test_models.py
```diff
@@ -21,6 +21,7 @@
 import shutil
 import tempfile
 from threading import Thread
+from collections import Counter, defaultdict
```
prathamesh920: Counter is not used!

yaksh/test_models.py
```diff
@@ -1553,6 +1554,20 @@ def test_duplicate_attempt_answerpaper(self):
...
+        all_questions = self.answerpaper.get_all_ordered_questions()
+        categorized_question_indices = \
+            self.answerpaper.get_categorized_question_indices()
+        questions = all_questions[0:4]
+        category_question_map = defaultdict(lambda: [])
+        for index, question in enumerate(questions, 1):
+            category_question_map[
+                    legend_display_types[question.type]["label"]
+                    ].append(index)
+        category_question_map = dict(category_question_map)
```
prathamesh920: Exactly same code is written as it is in the model method!
Remove this. Simply create a dict with key value expected and compare with the got dict.
Also will be good if there are more test cases for this.

yaksh/models.py
```diff
@@ -2086,6 +2098,15 @@ def is_attempt_inprogress(self):
     def get_previous_answers(self, question):
         return self.answers.filter(question=question).order_by('-id')
 
+    def get_categorized_question_indices(self):
+
+        category_question_map = defaultdict(lambda: [])
+        for index, question in enumerate(self.get_all_ordered_questions(), 1):
+            category_question_map[
+                legend_display_types[question.type]["label"]
```
prathamesh920: I don't see the use of defaultdict because ```lengend_display_types``` is a normal dict and used inside a default dict so will fail in case mormal dict has ```KeyError```.

yaksh/models.py
```diff
@@ -80,6 +81,17 @@
     ("exact", "Case Sensitive"),
     )
 
+legend_display_types = {
+        "mcq": {"category": "objective", "label": "Objective Type"},
```
prathamesh920: Nested dict not required.
Simply keep ```{"mcq": "Objective Type", ...}``` since name of the dict is enough to understand the intention.

ankitjavalkar: This is my opinion is cleaner and more configurable because the decision for the category name and label name for the same category can be made in one place instead of hard coding/if-else structure in the templates.

yaksh/models.py
prathamesh920: Yes agreed.

yaksh/models.py
prathamesh920: Yes but this ```category``` key is not used anywhere

CruiseDevice: @prathamesh920 removed the category key.

yaksh/test_models.py
```diff
@@ -1084,6 +1095,56 @@ def setUpClass(self):
...
+            user_ip=self.ip,
+            course=self.course
+        )
+
+        self.attempted_papers2 = AnswerPaper.objects.filter(
+            question_paper=self.question_paper2,
+            user=self.user
+        )
+
+        self.question_paper2.fixed_questions.add(*self.questions2)
```
prathamesh920: Why not first create question paper completely and then create answer paper?

yaksh/test_models.py
```diff
@@ -961,12 +966,17 @@ def setUpClass(self):
 
         all_questions = Question.objects.filter(user=self.user).order_by("id")
         self.questions = all_questions[0:3]
+        self.questions2 = all_questions[4:7]
+        self.questions3 = []
```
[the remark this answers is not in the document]
CruiseDevice: I wasn't able to guess any proper name and also noticed many other variables in the code named in the same format so named them question2 and question3. I guess reassigning won't be the problem.

yaksh/test_models.py
CruiseDevice: What difference does that make?

yaksh/test_models.py
```diff
@@ -1084,6 +1095,56 @@ def setUpClass(self):
...
+        )
+
+        self.attempted_papers2 = AnswerPaper.objects.filter(
+            question_paper=self.question_paper2,
+            user=self.user
+        )
+
+        self.question_paper2.fixed_questions.add(*self.questions2)
+        already_attempted2 = self.attempted_papers2.count()
+        self.answerpaper2.attempt_number = already_attempted2 + 1
```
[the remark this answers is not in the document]
CruiseDevice: @prathamesh920 I took the reference of the already written tests while writing these. I thought increasing attempt_number is required. But i tested it after removing the increment condition and it works fine. Just let me know if the increment condition is required or not. I will then make changes to the code.

yaksh/test_models.py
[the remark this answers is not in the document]
prathamesh920: If it's reassigned immediately below then your assignment is meaningless, remove it!

yaksh/test_models.py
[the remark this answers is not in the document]
prathamesh920: I still don't understand why that increment was done! The reference code must be doing something else.
If you don't understand any part of the code then ask Ankit or me before implementation.

yaksh/test_models.py
```diff
@@ -1084,6 +1095,56 @@ def setUpClass(self):
...
+        self.answerpaper2 = AnswerPaper(
+            user=self.user,
+            question_paper=self.question_paper2,
+            start_time=self.start_time,
+            end_time=self.end_time,
+            user_ip=self.ip,
+            course=self.course
+        )
+
+        self.attempted_papers2 = AnswerPaper.objects.filter(
```
prathamesh920: Why you need to do filter?
Already self.answerpaper2 gives you the required object!

yaksh/test_models.py
prathamesh920: It makes a difference.
You are creating question paper incompletely then creating answer paper with that incomplete question paper and saving it then,
you add questions to question paper.
Then add those questions to answer paper n save it!
They are behaving independently when they are related.
Also it increases complexity unnecessary.
Simply create a question paper completely, then u have make_answerpaper method of question paper to create answer paper. 
So simple and clear!

yaksh/test_models.py
```diff
@@ -952,7 +953,11 @@ def setUpClass(self):
         self.profile = self.user.profile
         self.quiz = Quiz.objects.get(description='demo quiz 1')
         self.question_paper = QuestionPaper(quiz=self.quiz, total_marks=3)
+        self.question_paper2 = QuestionPaper(quiz=self.quiz, total_marks=3)
+        self.question_paper3 = QuestionPaper(quiz=self.quiz, total_marks=3)
```
prathamesh920: Also check whether the marks are proper.
I mean marks=3 means that the question paper should create an answer paper with 3 marks total.

CruiseDevice:
> Can't you test with the existing answepaper?

Did same for the first test. Had to create new question_paper and answerpaper for other two tests.

yaksh/test_models.py
```diff
@@ -1105,6 +1106,36 @@ def setUpClass(self):
...
+        )
+        # create question_paper4
+        self.question_paper4 = QuestionPaper(
+            quiz=self.quiz, total_marks=3, shuffle_questions=True
+        )
+        self.question_paper4.save()
+        question_list_with_no_questions = []
+        self.question_paper4.fixed_questions.add(
+            *question_list_with_no_questions
+        )
```
prathamesh920: Can remove above two lines, also its understood that question paper is without any question.

yaksh/test_models.py
```diff
@@ -1105,6 +1106,36 @@ def setUpClass(self):
...
+        self.question_paper3.fixed_questions.add(
+            *question_list_with_only_one_category
+        )
+        # create anspaper for user1 with questions of only one category
+        self.user1_answerpaper2 = self.question_paper3.make_answerpaper(
+            self.user, self.ip, 1, self.course.id
+        )
+        # create question_paper4
+        self.question_paper4 = QuestionPaper(
+            quiz=self.quiz, total_marks=3, shuffle_questions=True
```
prathamesh920: total marks keep zero as this will be question paper with no questions

prathamesh920:
LGTM, can be merged.
We need to fix the PEP8 issue, can be done in a separate PR, but that should be ASAP!

CruiseDevice:
> LGTM, can be merged.
> We need to fix the PEP8 issue, can be done in a separate PR, but that should be ASAP!

They are already fixed in one of the PRs. Merging those will solve this.

ankitjavalkar:
@CruiseDevice Please make sure that Travis CI test build passes and I will merge this ASAP.

CruiseDevice:
@ankitjavalkar CI passed. Can merge it.

PR #537, opened 2019-01-09
[UI] Multiple UI Fixes in New UI

### LINKED ISSUE #536: [UI] Static 'Overall Course Completion' in Student Interface needs to display relevant info
- On the Page at "/exam/course_modules/<module_id>/" - 'Overall Course Completion' section is static, needs to display relevant info
- The page also directly uses {{ Course }}, need to replace this with {{ Course.name }}

ankitjavalkar:
- Deprecate usage of 'course_completion_percent' custom template tag
- Replace old code in course_detail.html
- Fix Overall Course Completion percentage in Student interface

Closes #536

### DIFF
--- a/yaksh/templates/yaksh/course_detail.html
+++ b/yaksh/templates/yaksh/course_detail.html
@@ -141,61 +141,47 @@
                         <thead>
                             <tr class="yakshlight">
                             <th>Sr No.</th>
-                            <th>Students</th>
-                            <th>Total</th>
+                            <th>Email</th>
+                            <th>Current Unit</th>
+                            <th>Course Completion Percentage</th>
                             <th>Grade</th>
-                            <th colspan="{{modules|length}}">Modules</th>
-                            </tr>
-                            <tr>
-                                <th scope="row"></th>
-                                <th></th>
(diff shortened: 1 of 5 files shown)


yaksh/templates/yaksh/course_detail.html
```diff
@@ -141,61 +141,47 @@
                         <thead>
                             <tr class="yakshlight">
                             <th>Sr No.</th>
-                            <th>Students</th>
-                            <th>Total</th>
+                            <th>Email</th>
```
prathamesh920: Below for Email column if email does not exists then name is displayed.
I would suggest that name and email to be separate columns.

ankitjavalkar: A User/Teaching assistant recommended that the Name or Email is not *as* import as the Roll number of the students.

I'll be including the Roll Number and keeping the rest of the structure unchanged as more columns will mean a crowded UI.

prathamesh920:
Slightly uncomfortable with full name shown in email column only if email does not exits.
Created issue #550 
I agree its not a major issue but still.
LGTM

PR #545, opened 2019-01-11
Fix Embed video link issue

### LINKED ISSUE #540: [UI] Embed video link issue
In Lesson form on click of Embed video link button, the popup is static and not at proper position.

CruiseDevice:
- Fixes Embed video link issue in LessonForm of Add New Lesson.
- It breaks because of compatibility issues between jQuery 3.3.1 and jQuery UI 1.9.1 .
  Using jQuery UI with the jQuery Migrate plugin fixes this issue.

Closes #540


ankitjavalkar:
> We can update our code to make sure it works with the newest jquery. This solution is a hack. @prathamesh920 Please comment.

This will have to do for now as porting the project JS components completely to a newer JQuery is time intensive and needs to be done with care.

@CruiseDevice Kindly create an issue mentioning that JQuery needs to be upgraded.

maheshgudi:
@ankitjavalkar @CruiseDevice Fine, approving this for now. But even the jquery-migrate team is very specific that it is to be used as a stopgap solution and not a permanent fix. Please make sure jquery-migrate is deprecated in the future.

adityacp:
I guess you need to add jquery ui 1.12 js link in add_module.html. Just check it once.

PR #551, opened 2019-01-21
Remove user as teacher of any course if is_moderator is set to False

### LINKED ISSUE #512: Resetting the is_moderator field to False should remove the user as Teacher from Courses

ankitjavalkar:
Closes #512


ankitjavalkar:
Test cases added

PR #552, opened 2019-01-23
Fix Embed video link issue and  Upgrade jQuery-ui.

### LINKED ISSUE #549: jQuery 3.3.1 breaks jQuery UI dialog with error jf.getClientRects is not a function
Temporary solution for this is to use JQuery UI with the jQuery Migrate plugin. As this seems to be a hack solution, it can be used for time until the jQuery ui is updated to the new compatible version. 

**This has been tested with following versions :**
jQuery: 3.3.1
jQuery UI: 1.9.1

CruiseDevice:
- Upgrade jQuery-ui.js and jQuery-ui.css from 1.9.1 to 1.12.1.
- Resolves jQuery breaking jQuery-ui.
- Fixes #549 and #540


ankitjavalkar:
@CruiseDevice I'll merge this as soon as you fix the conflict.

PR #566, opened 2019-02-01
Toggle view grade and other UI improvements

### LINKED ISSUE #560: Moderator should have the option to disable display Grade scored by students
Currently the Course Display Page within Student Interface directly displays the grade scored by Student for a particular course.

Moderator should be able to toggle this.

### LINKED ISSUE #569: Change Course Listing UI on Student Dashboard
Multiple fixes needed to the Course listing Page on Student Dashboard
- All courses sections should be clickable along with the DETAIL button
- Course completed status should be on first page of students dashboard
- Change button to continue if the course is left in between

ankitjavalkar:
Closes #569 
Closes #560

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -758,6 +758,7 @@ class Course(models.Model):
     teachers = models.ManyToManyField(User, related_name='teachers')
     is_trial = models.BooleanField(default=False)
     instructions = models.TextField(default=None, null=True, blank=True)
+    view_grade = models.BooleanField(default=False)
     learning_module = models.ManyToManyField(LearningModule,
                                              related_name='learning_module')
 
--- a/yaksh/templates/yaksh/quizzes_user.html
+++ b/yaksh/templates/yaksh/quizzes_user.html
@@ -72,21 +72,21 @@
         <div class="yakshwell">
             <div class="row yakshlabel align-items-center">
                 <div class="col">
-                    <h4><b>
-                        
-                        {{ course.name }} by {{ course.creator.get_full_name }}
-                        
-                    </b></h4>
-                    {% if course.is_active_enrollment %}
+                    <a data-toggle="collapse" href="#collapsedetails{{course.data.id}}" role="button" aria-expanded="false" aria-controls="#collapsedetails{{course.data.id}}">
+                        <h4><b>
+                            {{ course.data.name }} by {{ course.data.creator.get_full_name }}
+                        </b></h4>
+                    </a>
+                    {% if course.data.is_active_enrollment %}
                         <div class="text-left">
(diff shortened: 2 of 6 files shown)


yaksh/models.py
```diff
@@ -758,6 +758,7 @@ class Course(models.Model):
     teachers = models.ManyToManyField(User, related_name='teachers')
     is_trial = models.BooleanField(default=False)
     instructions = models.TextField(default=None, null=True, blank=True)
+    view_grade = models.BooleanField(default=False)
```
prathamesh920: Default should be True!
Coz False will affect the existing data.

yaksh/views.py
```diff
@@ -164,18 +165,27 @@ def quizlist_user(request, enrolled=None, msg=None):
...
     else:
         courses = Course.objects.filter(
             active=True, is_trial=False
         ).exclude(
            ~Q(requests=user), ~Q(rejected=user), hidden=True
-        )
+        ).order_by('-id')
         title = 'All Courses'
 
+    for c in courses:
```
prathamesh920: ```for course in courses:```

yaksh/models.py
ankitjavalkar: It will not, it is a completely new field and the field is only used for Frontend changes in Templates. Migrations will handle this.

yaksh/views.py
ankitjavalkar: fixed

ankitjavalkar:
@prathamesh920  Made changes and replied comments.

yaksh/models.py
prathamesh920: The value of view_grade for the existing Courses will be taken as False, right?
Then the students will not be able to see the grades that they were seeing, right?

ankitjavalkar: Yes, the only place where students are currently supposed to see grades in the live version is in Fellowship Course.

Basically, this is the same as Show AnswerPaper, we have to keep the default ```False``` since this is a Result, the decision to release the grade should be the Moderator's decision.

yaksh/templates/yaksh/quizzes_user.html
```diff
@@ -124,39 +131,55 @@ <h4><b>
...
+                        <div class="row my-3">
+                            <div class="col-md-5">
+                                ENDS ON
+                            </div>
+                            <div class="col-md-7">
+                                {{course.data.end_enroll_time}}
+                            </div>
+                        </div>
+                        <div class="row my-3">
+                                <div class="progress-bar bg-success" role="progressbar" style="width:{{course.completion_percentage}}%"; color: black;" aria-valuenow="100" aria-valuemin="0" aria-valuemax="100">{{course.completion_percentage}} % completed
```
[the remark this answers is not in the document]
ankitjavalkar: Cannot help in this case, if you observe, I need to dynamically change styling to change the bar according to percentage completed.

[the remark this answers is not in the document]
prathamesh920: Yes thought so.

PR #567, opened 2019-02-04
Long Course name and Module name breaks table layout

### LINKED ISSUE #568: Courses and Modules with long names does not display properly in Moderator Dashboard and in side navigation menu of some templates.

CruiseDevice:
Fixes #568


prathamesh920:
Can you please create an issue for this and refer here

yaksh/static/yaksh/css/yakshcustom.css
```diff
@@ -228,3 +228,50 @@ html {
...
+	text-align: left;
+}
+
+#moderator_table table th:first-child{
+	width:45%;
+}
+
+/* Monitor Table */
+.monitor_table table th:first-child{
+	width:40%;
```
prathamesh920: I completely disagree with this!
We do not need this custom CSS for simply truncating  characters.

My suggestions:
- On moderator landing page, remove button class for the Course link and this will be taken care of.
- Many places the text is overlapped by the other HTML element so simply modify the CSS class and it will be fine.

In short there is no need to truncate or create these custom CSS, simply fix the CSS classes used.

CruiseDevice: > * On moderator landing page, remove button class for the Course link and this will be taken care of.


@prathamesh920 I tried what you are suggesting. It didn't work, also in terms of responsiveness and beautiful website, keeping buttons as links were not appropriate. I tried making it more responsive and mobile friendly, you would know if you look at it. I don't know if you remember we discussed this issue and went forward with  your suggestion of avoiding `truncatechars` django filter and using `CSS` for solving this task. I've shown the result of removing buttons and keeping them just links to @ankitjavalkar . They looked horrible and were behaving same like the buttons were behaving.  After a discussing with him i moved on to using `CSS`



> * Many places the text is overlapped by the other HTML element so simply modify the CSS class and it will be fine.

I didn't find anywhere text getting overlapped by any other HTML Tag. 

It would be very helpful if you direct me to the code where you think this is happening.

As far as i think the only problem would be the naming convention of the css classes. I was thinking of using more generic name where they can be used on other places for truncating string if they are taking more place than required. 

Another add on would be to add `alt` as alternate text display onhover for displaying full text where the text is truncated for user to understand.

prathamesh920: 1.  As per our discussion I said no need of django filter and told you to make changes to the CSS.
2. The problem is with overlapping of HTML elements, you can find it if you do inspect elements.

Anyway kindly meet me now if you are free, will let you know what is expected.

CruiseDevice:
Not required anymore. Replaced by @adityacp PR

PR #575, opened 2019-02-12
Fix download file for a question in quiz

### LINKED ISSUE #580: Link of download file for a question is broken.
After clicking the link it does not download the file and redirects to the same page.

CruiseDevice:
- Display just the filename instead of complete path of the file on the server. 

Closes #580

### DIFF
--- a/yaksh/templates/yaksh/question.html
+++ b/yaksh/templates/yaksh/question.html
@@ -160,7 +160,7 @@ question_type = "{{ question.type }}"
                                 <span> Files to download for this question </span> <hr>
                                 {% for f_name in files %}
                                   <div class="yakshwell">
-                                    <a href="{{f_name.file.url}}" class="btn btn-outline-secondary" target="_blank"><b>{{forloop.counter}}.</b> {{f_name.get_filename}}</a>
+                                    <a href="{{f_name.file.url}}" class="btn btn-outline-secondary btn-sm " target="_blank">{{f_name.get_filename}}</a>
                                     <br>
                                     </div>
                                 {% endfor %}
--- a/yaksh/templates/yaksh/question.html
+++ b/yaksh/templates/yaksh/question.html
@@ -160,7 +160,7 @@ question_type = "{{ question.type }}"
                                 <span> Files to download for this question </span> <hr>
                                 {% for f_name in files %}
                                   <div class="yakshwell">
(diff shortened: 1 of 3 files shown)


yaksh/templates/yaksh/question.html
```diff
@@ -160,7 +160,7 @@
                                 <span> Files to download for this question </span> <hr>
                                 {% for f_name in files %}
                                   <div class="yakshwell">
-                                    <a href="{{f.file.url}}" class="btn btn-outline-secondary"><b>{{forloop.counter}}.</b> {{f_name.file.name}}</a>
+                                    <a href="{{f_name.file.url}}" class="btn btn-outline-secondary" target="_blank"><b>{{forloop.counter}}.</b> {{f_name.filename}}</a>
```
prathamesh920: It should be ```f_name.get_filename``` as you have added ```get_filename``` and not ```filename```
So kindly add a test case to avoid such error.

CruiseDevice:
Merging on priority

ankitjavalkar:
A minimal test case has been added for the method added in this changeset.

Merging on priority.

yaksh/test_models.py
```diff
@@ -2130,3 +2131,22 @@ def test_course_is_complete(self):
...
+        self.question = Question.objects.get(summary='Q1')
+        self.filename = "uploadtest.txt"
+        self.uploaded_file = SimpleUploadedFile(self.filename, b'Test File')
+        self.file_upload = FileUpload.objects.create(
+            file=self.uploaded_file,
+            question=self.question
+        )
+
+    def test_get_file_name(self):
+        print((self.file_upload.file.path))
```
adityacp: Remove print statement

yaksh/test_models.py
```diff
@@ -2130,3 +2131,22 @@ def test_course_is_complete(self):
...
+        self.assertEqual(self.file_upload.get_filename(), self.filename)
+
+    def tearDown(self):
+        os.remove(self.file_upload.file.path)
```
adityacp: check if the file exists before removing. It might happen sometimes that the file may not be present. Just a safety check.

ankitjavalkar: Fixed in subsequent PR #575

yaksh/test_models.py
ankitjavalkar: Fixed in subsequent PR #575

Issue #581, opened 2019-02-18
 [Production Issue] Fixes in new UI


adityacp:
Course Details Page UI issues:- 
- [x] Course details page does not have proper table alignment for enrolled students section. 
- [ ] Same issue as above in send mail section of the course details page.
- [x] Student name is shown in the `Current unit` column in view course status page.
- [x] No need of `Export to CSV` button in the course status page.
- [x] Hovering on the email in course status page does not show the tooltip properly.

Courses page UI issues:-
- [x] Teachers added to the course does not show properly. Checkbox alignment is not proper.

Login related UI fixes:-

- [ ] Change password is showing Django internal templates and not the override templates. Same for forgot password.
- [ ] After changing profile, reloading gif is not showing up.

Student Dashboard:- 

- [x] Unable to see download course button on student dashboard

CruiseDevice:
Adding screenshots related to the above issues here for reference:

* Student name is shown in the `Current unit` column in view course status page. [Screenshot below]

![image](https://user-images.githubusercontent.com/10258150/52993802-7b6d8280-343b-11e9-9da2-f52d324e03ee.png)

* Course details page does not have proper table alignment for enrolled students section [Screenshot below]

![image](https://user-images.githubusercontent.com/10258150/52993870-b374c580-343b-11e9-9269-468028d71fd0.png)

CruiseDevice:
> * Hovering on the email in course status page does not show the tooltip properly.

Already handled in Jquery update PR #552.

CruiseDevice:
> * Same issue as above in send mail section of the course details page.

@adityacp I think there is no problem with the send_email template. Can you please reproduce the error and post the screenshot here.

NeoHimu:
>     * Change password is showing Django internal templates and not the override templates. Same for forgot password.

unable to reproduce the issue.. working fine on my local system. @ankitjavalkar @prathamesh920

adityacp:
Fixed in #652

PR #585, opened 2019-02-20
Multiple UI Changes

CruiseDevice:
- Change course status page UI.
- Fix breaking alignment in Monitor page summary table.

Refer #581

### DIFF
--- a/yaksh/templates/yaksh/monitor.html
+++ b/yaksh/templates/yaksh/monitor.html
@@ -68,24 +68,39 @@ $(document).ready(function()
         {% if quiz %}
         {% if papers %}
         <div class="row">
-        <div class="card col-md-4">
-            <div class="row">
-                <div class="col"><b>Course Name:<br>Quiz Name:<br>Number of papers:<br>Papers completed:<br>Papers in progress:</b>
-                </div>
-                <div class="col">                
-                     {{ course.name }}<br>
-                     {{ quiz.description }}<br>
-                    {{ papers|length }}<br>
-                    {% completed papers as completed_papers %}
-                     {# template tag used to get the count of completed papers #}
-                      <b class="yakshred"> {{ completed_papers }} </b><br>
-                    {% inprogress papers as inprogress_papers %}
-                     {# template tag used to get the count of inprogress papers #}
-                     <b class="yakshred">  {{ inprogress_papers }} </b>
-                </div>
+        <div class="card col-md-10">
+            <div class = "table-responsive"">
+                <table class = "table">
+                    <tr>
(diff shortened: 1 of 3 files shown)


yaksh/templates/yaksh/monitor.html
```diff
@@ -68,24 +68,39 @@
...
-                     {{ quiz.description }}<br>
-                    {{ papers|length }}<br>
-                    {% completed papers as completed_papers %}
-                     {# template tag used to get the count of completed papers #}
-                      <b class="yakshred"> {{ completed_papers }} </b><br>
-                    {% inprogress papers as inprogress_papers %}
-                     {# template tag used to get the count of inprogress papers #}
-                     <b class="yakshred">  {{ inprogress_papers }} </b>
-                </div>
+        <div class="card col-md-10">
```
prathamesh920: This border does not align with the result table below.
The result table is slightly shifted towards the right.
Also these two tables are touching each other which doesn't look good.

yaksh/templates/yaksh/monitor.html
```diff
@@ -68,24 +68,39 @@
...
+                        <td>
+                            {% completed papers as completed_papers %}
+                            <b class = "yakshred">{{completed_papers}}</b>
+                        </td>
+                    </tr>
+                    <tr>
+                        <td><b>Papers in progress: &nbsp</b></td>
+                        <td>
+                            {% inprogress papers as inprogress_papers %}
+                            <b class="yakshred">  {{ inprogress_papers }} </b>
```
prathamesh920: since you are modifying this can we have different color and not same as above.
My suggestion completed can be green and in-progress can be red.

yaksh/templates/yaksh/monitor.html
CruiseDevice: > Also these two tables are touching each other which doesn't look good.

Can you please clear you chache and check once again. Also try this with the updated PR

yaksh/templates/yaksh/monitor.html
```diff
@@ -68,32 +68,49 @@
...
+                    </tr>
+                    <tr>
+                        <td><b>Number of papers: &nbsp</b></td>
+                        <td>{{papers|length}}</td>
+                    </tr>
+                    <tr>
+                        <td><b>Papers Completed: &nbsp</b></td>
+                        <td>
+                            {% completed papers as completed_papers %}
+                            <b class = "yakshgreen">{{completed_papers}}</b>
```
prathamesh920: Next time please see to it that the code writing style is consistent with the current code base.
There shouldn't be spaces around = sign.
Merging for now.

PR #587, opened 2019-02-21
Add Download Course button in Course Module template

CruiseDevice:
- Add Download Course button in Course Module template.

Refer #581

### DIFF
--- a/yaksh/templates/yaksh/course_modules.html
+++ b/yaksh/templates/yaksh/course_modules.html
@@ -20,7 +20,7 @@
             <div class="col h4">  {{ course.name }}  </div>
             <div class = "col">
                 {% if course.has_lessons %}
-                <a href="{{URL_ROOT}}/exam/download_course/{{course.id}}" data-toggle="tooltip" title="Download course content" class="btn btn-primary pull-right">
+                <a href="{% url 'yaksh:download_course' course.id %}" data-toggle="tooltip" title="Download course content" class="btn btn-primary pull-right">
                     Download Course
                 </a>
                 {% endif %}
--- a/yaksh/templates/yaksh/course_modules.html
+++ b/yaksh/templates/yaksh/course_modules.html
@@ -14,10 +14,17 @@
 {% block main %}
 
 <div class="container">
-<div class="row justify-content-md-center yakshwell ">
-    <div class="col-md-10 bg-light card">
-        <div class="row align-items-center my-3">
-            <div class="col h4 text-center">  {{ course.name }}  </div>
+<div class="row justify-content-md-center yakshwell">
+    <div class="col-md-10 bg-light card" id="card">
+        <div class="row align-items-center">
+            <div class="col h4">  {{ course.name }}  </div>
+            <div class = "col">
+                {% if course.has_lessons %}
+                <a href="{{URL_ROOT}}/exam/download_course/{{course.id}}" data-toggle="tooltip" title="Download course content" class="btn btn-primary pull-right">
+                    Download Course
+                </a>
+                {% endif %}
+            </div>
         </div>
     </div>
 </div>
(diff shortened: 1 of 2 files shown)


yaksh/templates/yaksh/course_modules.html
```diff
@@ -15,22 +15,17 @@
 {% block main %}
 
 <div class="container">
-<div class="row justify-content-md-center yakshwell ">
+<div class="row justify-content-md-center yakshwell">
     <div class="col-md-10 bg-light card">
-        <div class="row align-items-center">
+        <div class="row align-items-center" style="height: 100px;">
```
prathamesh920: Kindly avoid such inline CSS!

yaksh/templates/yaksh/course_modules.html
```diff
@@ -15,22 +15,17 @@
...
-                    </div>
-                    <div class="col-sm-5">
-                    <span class="progress">
-                      <span class="progress-bar bg-warning " role="progressbar" style="width: {{ course_percentage }}%; color: black;" aria-valuenow="50" aria-valuemin="0" aria-valuemax="100">{{ course_percentage }} % completed</span>
-                    </span>
-                    </div>
-                </div>
-            </div>   
+            <div class = "col">
+                {% if course.has_lessons %}&nbsp;&nbsp;&nbsp;
```
prathamesh920: Avoid the use of so many```&nbsp;``` !
Use proper divs or HTML elements or CSS property instead.

yaksh/templates/yaksh/course_modules.html
```diff
@@ -15,22 +15,17 @@
...
+                <a href="{{URL_ROOT}}/exam/download_course/{{course.id}}" data-toggle="tooltip" title="Download course content" class="btn btn-primary pull-right">
```
prathamesh920: Can we start using ```url``` template tag instead of given path?

yaksh/templates/yaksh/course_modules.html
CruiseDevice: @prathamesh920  Will it be okay to create a new class selector just for one tag in css file?

yaksh/templates/yaksh/course_modules.html
CruiseDevice: @prathamesh920 Will it be okay to create a new class selector just for one tag in css file?

yaksh/templates/yaksh/course_modules.html
prathamesh920: Firstly, try to achieve it by proper use of classes provided by the bootstrap CSS.

yaksh/templates/yaksh/course_modules.html
prathamesh920: same as above.
Also, use proper divs or HTML elements or CSS property instead

yaksh/templates/yaksh/course_modules.html
CruiseDevice: @prathamesh920 For some reason django url tag is causing problem. It throws NoReverseMatch error.

prathamesh920: Can you share your code and error with me

CruiseDevice: @prathamesh920 
```
<a href="{% url 'download_course' course.id %}" data-toggle="tooltip" title="Download course content" class="btn btn-primary pull-right">Download Course</a>
```

prathamesh920: Use namespace yaksh.
So in tag call ```'yaksh:download _coursr'```
Above should work.

CruiseDevice: Usually namespaces are used in apps urls.py as global variable. I didn't see that in root urls.py

prathamesh920:
In description "Remove redundant Course completion status" is mentioned.
I think this PR doesn't do that.
Also, add issue reference.

PR #591, opened 2019-03-05
Update Login UI 

NeoHimu:
basic cleanup which includes removing history and ad informations from the login page


prathamesh920:
Please refer the issue number here.

prathamesh920:
Also, the login page is the landing page for all the users and the write up is missing.
I think we should have a small write up for Yaksh.

ankitjavalkar:
> Also, the login page is the landing page for all the users and the write up is missing.
I think we should have a small write up for Yaksh.

@prathamesh920 Maybe, but this was done after feedback from @prabhuramachandran 
In order to keep the application generic, the page can be kept blank. Otherwise, @NeoHimu please create an issue for this and kindly check what content needs to be added and add it to this PR

ankitjavalkar:
@NeoHimu the migration file should not have been committed in the first place. Kindly undo the latest commit, remove the migration file and use ```commit --amend``` to ensure that there is only one commit.

ankitjavalkar:
Closing the PR as the migration files were not completely erased from commits as advised.

PR #592, opened 2019-03-12
Add functionality to restrict access to next module based on passing previous module

### LINKED ISSUE #597: Functionality to restrict access to next module based on passing previous module

ankitjavalkar:
@prathamesh920 @adityacp @CruiseDevice Please review, will fix tests on an ongoing basis and keep updating the PR

Closes #597

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -716,17 +716,17 @@ class LearningModule(models.Model):
 
     def get_passing_status(self, user, course):
         course_status = CourseStatus.objects.filter(user=user, course=course)
-        state = "failed"
         if course_status.exists():
-            ordered_units = self.learning_unit.filter(type='quiz').order_by("order")
+            learning_units_with_quiz = self.learning_unit.filter(type='quiz')
+            ordered_units = learning_units_with_quiz.order_by("order")
 
-        status_list = [unit.quiz.get_answerpaper_passing_status(user, course)
+        statuses = [unit.quiz.get_answerpaper_passing_status(user, course)
                        for unit in ordered_units]
 
-        if not status_list:
+        if not statuses:
             status = False
         else:
-            status = all(status_list)
+            status = all(statuses)
         return status
 
     def is_prerequisite_passed(self, user, course):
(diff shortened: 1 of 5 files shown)


yaksh/models.py
```diff
@@ -539,6 +539,18 @@ def get_answerpaper_status(self, user, course):
...
         return status
 
+    def get_answerpaper_passing_status(self, user, course):
+        try:
+            qp = self.questionpaper_set.get().id
+        except QuestionPaper.DoesNotExist:
+            qp = None
+        ans_ppr = AnswerPaper.objects.filter(
+            user=user, course=course, question_paper=qp
+        ).order_by("-attempt_number")
```
prathamesh920: Should this be order by percentage?
As there can be a case that in 2 attempts, a user passed in the first and failed in the second!
What are your thoughts?

yaksh/models.py
```diff
@@ -698,6 +714,33 @@ def is_prerequisite_passed(self, user, course):
                 success = False
         return success
 
+    def get_passing_status(self, user, course):
+        course_status = CourseStatus.objects.filter(user=user, course=course)
+        state = "failed"
```
prathamesh920: This variable is not used.

yaksh/models.py
```diff
@@ -698,6 +714,33 @@ def is_prerequisite_passed(self, user, course):
...
+        if course_status.exists():
+            ordered_units = self.learning_unit.filter(type='quiz').order_by("order")
+
+        status_list = [unit.quiz.get_answerpaper_passing_status(user, course)
```
prathamesh920: Can we have a name like ```all_status``` or  may be ```statuses```?

prathamesh920:
Also fix PEP8

yaksh/models.py
ankitjavalkar: Done

yaksh/models.py
ankitjavalkar: Removed

ankitjavalkar:
Fixed PEP8

ankitjavalkar:
@prathamesh920 Please check. PEP8 issues have been fixed, but PEP8 issue comment is now outdated.

PR #596, opened 2019-04-05
Change in_dir directory nomenclature for code server

### LINKED ISSUE #594: Special unicode characters in Usernames prevents successful evaluation
User can signup with usernames containing special unicode characters like Î or Ä, this causes a failure on the backend in the code_server when it creates a directory because unicode characters are not supported.

Error Traceback:
```
[root@ee09a50a8361 src]# tailf /Temp/test.log 
    self._target(*self._args, **self._kwargs)
  File "/src/yaksh/code_server.py", line 55, in check_code
    result = grader.evaluate(data)
  File "/src/yaksh/grader.py", line 99, in evaluate
    self.setup()
  File "/src/yaksh/grader.py", line 111, in setup
    if not os.path.exists(self.in_dir):
  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/genericpath.py", line 19, in exists
    os.stat(path)
UnicodeEncodeError: 'ascii' codec can't encode character '\xee' in position 55: ordinal not in range(128)
```

ankitjavalkar:
Closes #594

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -1103,7 +1103,7 @@ class Profile(models.Model):
     def get_user_dir(self):
         """Return the output directory for the user."""
 
-        user_dir = join(settings.OUTPUT_DIR, str(self.user.username))
+        user_dir = join(settings.OUTPUT_DIR, str(self.user.id))
         if not exists(user_dir):
             os.makedirs(user_dir)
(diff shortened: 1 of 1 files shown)


ankitjavalkar:
@adityacp Please comment if you think this change affects other parts like upload file/code applied on uploaded files.

Thanks

PR #605, opened 2019-06-24
Permisson manager

dshubhadeep:


permissions/migrations/0006_permission_course.py
```diff
@@ -0,0 +1,24 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='permission',
+            name='course',
+            field=models.ForeignKey(
+                null=True,
+                on_delete=django.db.models.deletion.CASCADE,
+                to='yaksh.Course'),
+        ),
+    ]
```
adityacp: You don't need to push migrations file. Please remove it.

permissions/static/js/team.js
```diff
@@ -0,0 +1,42 @@
...
+    url: "/permissions/units/get/",
+    success: function(result) {
+      addModuleOptions(result);
+    }
+  });
+});
+
+function addModuleOptions({ units }) {
+  let html = `
+    <div class="form-group" id="unit-group">
```
adityacp: Instead of using direct HTML tags in JS. You can use document.createElement(). For e.g document.createElement("div");

permissions/utils.py
```diff
@@ -0,0 +1,43 @@
+def check_permission(course, unit, user):
+    ''' Check if user has permissions for a given unit '''
+
+    teams = list(course.team_set.all())
```
adityacp: you don't need to typecast to list. Also do you need all the columns from team set?
You can use `values_list` to just get the ids and use it in the next filter.

permissions/utils.py
```diff
@@ -0,0 +1,43 @@
...
+
+            object_perms = list(filter(
+                lambda x: x.content_object == unit,
+                permissions
+            ))
+
+            if len(object_perms):
+                return object_perms[0]
+
+    return None
```
adityacp: you don't need to do return None.

permissions/views.py
```diff
@@ -0,0 +1,206 @@
...
+
+# Create your views here.
+
+
+@login_required
+def home(request):
+    # Get users
+    users = User.objects.all().exclude(username=request.user.username)
+    teams = request.user.team_members.all()
+    courses = Course.objects.filter(creator=request.user)
```
adityacp: Don't use request.user everytime instead store it in a variable.

permissions/views.py
```diff
@@ -0,0 +1,206 @@
...
+        team = Team.objects.create(
+            name=team_name,
+            created_by=request.user
+        )
+
+        for member in members_list:
+            team.members.add(User.objects.get(username=member))
+
+        for course_id in courses_list:
+            team.courses.add(Course.objects.get(pk=course_id))
```
adityacp: No need of using for loop instead make a queryset of users and courses and add it to the team at once.

permissions/views.py
```diff
@@ -0,0 +1,206 @@
...
+
+        team.save()
+
+    return redirect("permissions:home")
+
+
+@login_required()
+def team_detail(request, team_id):
+    print("REQUEST PATH", request.path)
+    print("REQUEST INFO", request.path_info)
```
adityacp: Remove print statements

permissions/views.py
```diff
@@ -0,0 +1,206 @@
...
+
+        for member in team.members.all():
+            member_roles = list(
+                map(lambda x: x.name, member.role_set.filter(team=team)))
+            role_map[member.username] = ",".join(member_roles)
+
+        context["role_map"] = role_map
+
+    except Team.DoesNotExist:
+        print("Team doesn't exist")
```
adityacp: Remove print statement.

permissions/views.py
```diff
@@ -0,0 +1,206 @@
...
+        role_map = {}
+
+        for member in team.members.all():
+            member_roles = list(
+                map(lambda x: x.name, member.role_set.filter(team=team)))
+            role_map[member.username] = ",".join(member_roles)
+
+        context["role_map"] = role_map
+
+    except Team.DoesNotExist:
```
adityacp: Don't use try except for checking if Team exists, instead use exists().

permissions/views.py
```diff
@@ -0,0 +1,206 @@
...
+            team=team
+        )
+
+        role.save()
+
+        for member in members:
+            role.members.add(User.objects.get(username=member))
+
+        role.save()
+    except Team.DoesNotExist:
```
adityacp: Same here. Use exists().

permissions/views.py
```diff
@@ -0,0 +1,206 @@
...
+    for module in modules:
+        learning_units = module.learning_unit.all()
+        for learning_unit in learning_units:
+            if learning_unit.type == "quiz":
+                units.append({"key": "quiz_{}".format(learning_unit.quiz.id),
+                              "name": learning_unit.quiz.description})
+            else:
+                units.append(
+                    {"key": "lesson_{}".format(learning_unit.lesson.id),
+                     "name": learning_unit.lesson.name})
```
adityacp: I guess we already have a model method to get all the learning units for a module. Maybe you can use that.

adityacp:
Permissions is a different app but it requires course as a foreign key. I feel that should not be the case. It will be restricted to course only. It should be independent of course so that it can be reused in any other place.

permissions/static/js/team.js
dshubhadeep: This can be done, but I feel this makes the code more verbose without any benefits. The current method is pretty readable and easy to modify.

adityacp: Instead of you making template in JS. You can create a template and load the data in the template in views and send to your main template.
Using html tags directly is not a cleaner way of doing.

permissions/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
+
+# Create your tests here.
```
adityacp: It would be good if you add some tests in the app.

dshubhadeep: Yeah, I'm working on it.

permissions/views.py
```diff
@@ -0,0 +1,204 @@
...
+
+@require_POST
+@login_required
+def create_team(request):
+    user = request.user
+    team_name = request.POST.get("team_name")
+    members_list = request.POST.getlist("members")
+    courses_list = request.POST.getlist("courses")
+
+    if len(team_name) > 0:
```
adityacp: Do you need `len`? Instead you can directly do `if team_name` will also check for None.

permissions/views.py
```diff
@@ -0,0 +1,204 @@
...
+        )
+
+        members = User.objects.filter(username__in=members_list)
+        team.members.add(*members)
+        team.members.add(user)
+
+        courses = Course.objects.filter(id__in=courses_list)
+        team.courses.add(*courses)
+
+        team.save()
```
adityacp: Do you need this? Just doing `add` should save the data in the database.

permissions/views.py
dshubhadeep: Yeah, that should also work

permissions/views.py
```diff
@@ -0,0 +1,204 @@
...
+            course=course,
+            perm_type=perm_type,
+            content_object=content_object
+        )
+
+        permission.save()
+
+        permission.role.add(role)
+
+        permission.save()
```
adityacp: This is redundant.

permissions/views.py
```diff
@@ -0,0 +1,204 @@
...
+    '''
+    Delete permission.
+    Only team creator can delete
+    '''
+
+    try:
+        team = Team.objects.get(pk=team_id)
+
+        if team.created_by == request.user:
+            Permission.objects.get(pk=permission_id).delete()
```
adityacp: Will this work? I guess delete works on queryset and not on object.

permissions/views.py
dshubhadeep: Yeah, just `add` saves it to the db.

permissions/views.py
dshubhadeep: Yeah, this works. [reference](https://stackoverflow.com/questions/3805958/how-to-delete-a-record-in-django-models)

permissions/views.py
```diff
@@ -0,0 +1,204 @@
...
+def delete_permission(request, permission_id, team_id):
+    '''
+    Delete permission.
+    Only team creator can delete
+    '''
+
+    try:
+        team = Team.objects.get(pk=team_id)
+
+        if team.created_by == request.user:
```
adityacp: You don't need this if you just get team by id and user.

permissions/views.py
```diff
@@ -0,0 +1,204 @@
...
+            {"key": "quiz_{}".format(quiz_unit.id),
+             "name": quiz_unit.description} for quiz_unit in quiz_units
+        ]
+
+        lesson_units_data = [
+            {"key": "lesson_{}".format(lesson_unit.id),
+             "name": lesson_unit.name} for lesson_unit in lesson_units
+        ]
+
+        units = quiz_units_data + lesson_units_data
```
adityacp: Do you need this extra list quiz_units_data and lesson_units_data? you can directly send all the units.

dshubhadeep: Yeah, I need the formatted key, so that i can refer what type of unit it is in the frontend. 

> Also instead of + you can use append

I'll change this.

adityacp: Yeah, I need the formatted key, so that i can refer what type of unit it is in the frontend
- In that case create a model method which returns specific data that you need. Don't get unnecessary data from database if you are not going to use it.

+ works fine. Is there any reason not to use it ?
- Sorry I got it wrong instead of `append` it is `extend`. `+` is slow as compared to append/extend.

ankitjavalkar: > create a model method

I do not think that the format of data used in the templates should drive the model methods, model methods should be driven by the API that you think that the application should have.

Anyway, in that case you can avoid ```get_lesson_units```and ```get_quiz_units``` and directly query modules (using the ```get_all_units``` method) and fetch desired values only using values_list

Once changed you will not need to use extend or concatenation of any kind.

yaksh/views.py
```diff
@@ -1,2955 +1,3004 @@
...
-        course = get_object_or_404(Course, id=course_id)
-        if not course.is_creator(current_user) and not \
-                course.is_teacher(current_user):
-            raise Http404('This course does not belong to you')
-
-        has_quiz_assignments = AssignmentUpload.objects.filter(
-                                question_paper_id=questionpaper_id
-                                ).exists()
-        context = {
-            "users"
```
ankitjavalkar: Please check why this humongous blob of change is showing up. Please undo and remove if this seems to be an error.

ankitjavalkar:
Pushed to the main repository as a branch for posterity.

PR #606, opened 2019-06-24
Selenium tests

techfreakology:


yaksh/live_server_tests/selenium_test.py
```diff
@@ -91,72 +156,368 @@ def test_c_question(self, question_label):
...
+    def check_enroll(self, username):
+        # Check whether the student is enrolled or not
+        WebDriverWait(
+            self.driver, 10).until(
+            EC.presence_of_element_located(
+                (By.XPATH, "//*[@id='enrolled_table']/tbody/tr/td[3]")))
+        soup = BeautifulSoup(self.driver.page_source, features="lxml")
+        cell = soup.find("table").find(
+            "tbody").find_all("tr")[0].find_all("td")[2]
+        return (cell.get_text().lower().strip() == username)
```
adityacp: Do you need beautiful soup for parsing through table? I guess Selenium can do it for you,

techfreakology: Done. Removed beautiful soup and now using selenium only to get the text.

yaksh/live_server_tests/selenium_test.py
```diff
@@ -91,72 +155,365 @@ def test_c_question(self, question_label):
...
+        ).click()
+
+    def open_questions(self):
+        # Open questions page
+        self.driver.find_element_by_link_text("Questions").click()
+
+        test_file = open(os.path.join(TEST_DIR, 'test_questions.json'), 'r')
+        test_questions = json.loads(test_file.read())
+        test_file.close()
+        id = 1
```
adityacp: You don't need id you can directly use `enumerate`

yaksh/live_server_tests/selenium_test.py
```diff
@@ -91,72 +155,365 @@ def test_c_question(self, question_label):
...
+
+        # Fill snippet
+        snippet = self.driver.find_element_by_id("id_snippet")
+        snippet.send_keys(question.snippet)
+
+        # Set Minimum Time
+        min_time = self.driver.find_element_by_id("id_min_time")
+        min_time.send_keys(question.min_time)
+
+        id = 0
```
adityacp: Same here. Use `enumerate`

yaksh/live_server_tests/load_test.py
```diff
@@ -31,8 +31,11 @@ def setUpClass(cls):
         app_label = 'yaksh'
         group_name = 'moderator'
         cls.group = create_group(group_name, app_label)
```
adityacp: You can create moderator group and add to it instead of using create_group method.

adityacp:
Remove prints from your code.
Also it would be nice if you add command to run tests in .travis.yml file.
You can check the existing  commands in travis file.

yaksh/live_server_tests/load_test.py
techfreakology: https://github.com/FOSSEE/online_test/blob/519cbb803c8020527578d2e9cc1b55a20f29c8ee/yaksh/live_server_tests/load_test.py#L33 
I have picked this way from the original repo, by using this method I don't have to rewrite all the code to give permissions.

techfreakology:
> Remove prints from your code.

The prints in the files are for providing verification that the tasks are performed successfully by selenium. I can remove them if you want.

> Also it would be nice if you add command to run tests in .travis.yml file. You can check the existing commands in travis file.

I have replaced previous load_test.py file and the command for executing this file is already present in .travis.yml file.

yaksh/live_server_tests/load_test.py
adityacp: It should not be dependent on a management method. If the method is removed it will fail the tests unnecessarily. Better is it should be independent

yaksh/live_server_tests/load_test.py
```diff
@@ -30,7 +30,20 @@ def setUpClass(cls):
 
         app_label = 'yaksh'
         group_name = 'moderator'
-        cls.group = create_group(group_name, app_label)
+
+        try:
+            cls.group = Group.objects.get(name=group_name)
+        except Group.DoesNotExist:
+            cls.group = Group(name=group_name)
+            cls.group.save()
```
adityacp: You don't need try and except in this case. instead of `get` use `get_or_create` and remove try except.

yaksh/live_server_tests/load_test.py
```diff
@@ -30,7 +30,20 @@ def setUpClass(cls):
...
+            # Get the models for the given app
+            content_types = ContentType.objects.filter(app_label=app_label)
+            # Get list of permissions for the models
+            permission_list = Permission.objects.filter(
+                content_type__in=content_types)
+            cls.group.permissions.add(*permission_list)
+            cls.group.save()
```
adityacp: you don't need to do this.

ankitjavalkar:
Pushed to the main repository as a branch for posterity.

PR #611, opened 2019-06-28
Corrected Python versioning

pllvsis:
Various versions of Python were used across the file. This update unifies the version used as per the user's version of Python.


tasks.py
```diff
@@ -17,6 +17,7 @@
 CHECK_FILE_PATH = os.path.join(SCRIPT_DIR, 'yaksh_data', CHECK_FILE)
 OS_NAME = sys.platform
 
+python_version = "python" if sys.version_info[0] < 3 else "python3"
```
ankitjavalkar: Please rename this like you would a constant, in all caps.

ankitjavalkar:
Thanks for sending in a PR. Please use git squash to squash both the commits into one single commit, TBH a merge operation was not entirely necessary.

tasks.py
```diff
@@ -218,28 +219,28 @@ def deploy(ctx, fixtures=False, static=True):
...
+            + python_version + " manage.py loaddata demo_fixtures.json"
         cmd = get_cmd(run_as_cmd, base_fixture_cmd)
         print("** Loading fixtures into database **")
         ctx.run(cmd)
         print("** Done! Loaded fixtures into database **")
 
     if static:
         base_static_cmd = "docker exec -i yaksh_django " \
-            "python3 manage.py collectstatic"
+            + python_version + " manage.py collectstatic"
```
adityacp: In every place instead of using `+` use string `format` method.

PR #622, opened 2019-07-03
Download course from yaksh for offline purpose.

CruiseDevice:


yaksh/models.py
```diff
@@ -1042,6 +1068,7 @@ def is_student(self, user):
     def create_zip(self, path, static_files):
         zip_file_name = string_io()
         with zipfile.ZipFile(zip_file_name, "a") as zip_file:
+            # print("HELLOOOOOOO - - - -- - - -- ")
```
prathamesh920: Please remove.

yaksh/static/yaksh/css/offline.css
```diff
@@ -1,18 +1,92 @@
...
+section{
+  padding: 2rem;
+}
+
+.bg-grey{
+  background-color:#dbf9f2;
+}
+
+
+/*.footer {
```
prathamesh920: Can remove this

yaksh/templates/yaksh/quiz.html
```diff
@@ -0,0 +1,114 @@
...
+                </center>
+                {% block content %}
+                    <center><h2>{{module.name}}</h2></center>
+                    <hr>
+                    <div id="status"></div>
+                    <hr>
+                    <br>
+                    <div class="text-center">
+                        <!-- Change Yaksh TODO -->
+                        <a class="btn btn-success" href="http://127.0.0.1:8000/exam/start/{{quiz.questionpaper_set.get.id}}/{{module.id}}/{{course.id}}" target="_blank">{{quiz.description}}</a>
```
prathamesh920: This should be yaksh.fossee.in instead of 127.0.0.1

prathamesh920:
LGTM, just a few suggestions
Better if the Download Course button is displayed with the course on the dashboard along with Details button.
The offline templates are with the other templates, we can add a folder "offline" and put those templates there.

CruiseDevice:
> LGTM, just a few suggestions
> Better if the Download Course button is displayed with the course on the dashboard along with Details button.
> The offline templates are with the other templates, we can add a folder "offline" and put those templates there.

Why not do this in a new issue rather than doing it all at once in this issue?

prathamesh920:
Okay please create an issue and refer here as to do later.
Also, the Download button display along with Details button  can be done now as it seems important from UX perspective.

CruiseDevice:
#623 TODO

ankitjavalkar:
> Also, the Download button display along with Details button can be done now as it seems important from UX perspective.

@prathamesh920 When the user is on this page, he may not be enrolled in the course yet, that is why, the button is kept one level inside. I think it should stay there, instead of doing if-else on the main page.

Kindly comment and resolve this ASAP.

prathamesh920:
I know that the user may not be enrolled but then in the download course code we check if the user is enrolled or not and then only he can download.
So I think no harm in giving the button there or with the courses under enrolled courses tab.
Infact will be better for the users.

prathamesh920:
Can be merged

Issue #625, opened 2019-07-30
Course detail page should have username column in the table.


CruiseDevice:
![image](https://user-images.githubusercontent.com/10258150/62102441-94197880-b2b6-11e9-976a-4534ab9466f8.png)

Or a better idea could be to make Full Name Column in course detail to redirect to user profile. User profile should also have an username section.

PR #629, opened 2019-09-16
Link User full name to user profile in course detail

CruiseDevice:
- Makes it easy to get user details.

Refer #625

### DIFF
--- a/yaksh/templates/yaksh/course_detail.html
+++ b/yaksh/templates/yaksh/course_detail.html
@@ -104,7 +104,7 @@
                                             <tr>
                                             <td><input type="checkbox" name="check" value="{{ enrolled.id }}"></td>
                                             <td>{{ forloop.counter }}.</td>
-                                            <td> {{ enrolled.get_full_name|title }} </td>
+                                            <td> <a href="{% url 'yaksh:view_profile' enrolled.id %}">{{ enrolled.get_full_name|title }}</a></td>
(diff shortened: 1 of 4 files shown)


yaksh/views.py
```diff
@@ -1710,9 +1710,12 @@ def grade_user(request, quiz_id=None, user_id=None, attempt_number=None,
 @login_required
 @has_profile
 @email_verified
-def view_profile(request):
+def view_profile(request, user_id=None):
     """ view moderators and users profile """
-    user = request.user
+    if user_id:
```
ankitjavalkar: Since any body can brute their way and check other peoples profiles, which was not originally intended, please add the following,

- Add a check to make sure that only Moderators can see other peoples profile.
- Query should be made with username instead of user_ids

yaksh/templates/yaksh/course_detail.html
```diff
@@ -272,7 +272,8 @@ <h5>Course Status</h5>
                                             <tr>
                                             <td><input type="checkbox" name="check" value="{{ enrolled.id }}"></td>
                                             <td>{{ forloop.counter }}.</td>
-                                            <td> {{ enrolled.get_full_name|title }} </td>
+                                            {{enrolled.id}}
```
ankitjavalkar: Why is this stray info here, should it not be inside a <td> tag?

PR #523, opened 2018-08-23
Add syntax highlight for code answers in answerpaper

maheshgudi:
Adds syntax highlight to answers displayed in grade_user, user_data and view_answerpaper.

PR #633, opened 2019-12-18
Add syntax highlight for code answers in answerpaper

ankitjavalkar:
Fixed conflicts in #523 and submitted fresh PR

@adityacp @CruiseDevice Kindly review.

### DIFF
--- a/requirements/requirements-common.txt
+++ b/requirements/requirements-common.txt
@@ -9,3 +9,4 @@ selenium==2.53.6
 coverage
 ruamel.yaml==0.15.23
 markdown==2.6.9
+pygments==2.2.0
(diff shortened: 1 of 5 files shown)


CruiseDevice:
LGTM

adityacp:
LGTM

maheshgudi:
@ankitjavalkar @adityacp While at the time I thought this was a good idea, in retrospect though, I think this is somewhat expensive to be done from the backend.  Maybe doing it using codemirror would be far less intensive. Just a suggestion though, please consider it if you can.

adityacp:
I feel it is ok to have it on the server side for now. We can check it on the production, if it makes lot of difference in the response time then we can add this to the client side.

PR #647, opened 2020-02-10
Register models in admin and add filters

adityacp:
- Register course status model in admin
- Add object str name in CourseStatus model
- Add filter in the admin for filtering courses


yaksh/admin.py
```diff
@@ -14,12 +16,36 @@ class ProfileAdmin(admin.ModelAdmin):
...
+class LearningModuleAdmin(admin.ModelAdmin):
+    list_filter = ['active', 'is_trial']
+
+
+class LessonAdmin(admin.ModelAdmin):
+    list_filter = ['active']
+
+
+class QuizAdmin(admin.ModelAdmin):
+	list_filter = ['active', 'is_trial']
```
ankitjavalkar: Quickfix: Looks like a tab?

PR #655, opened 2020-03-04
Bump django version to 3.0.3

adityacp:


adityacp:
Did you update the Django 1.11 to 3.0.3? Paginator `get_page` function is new in Django 3. It is not available in Django 1.11.

CruiseDevice:
Updating python version to 3.6 and then installing the Django 3.0.3 fixes this issue.

CruiseDevice:
LGTM

ankitjavalkar:
* All references to the ```six``` module need to be removed.
* We have used encode/decode multiple times to be able to support both Py2 & Py3, that needs to go as well.

adityacp:
I have removed the `six` module. encode/decode was used specifically for python 3. Removing encode/decode causes TypeError. So  I feel it is okay to keep it as it is.

PR #660, opened 2020-03-09
Add a button to download course progress in a CSV

### LINKED ISSUE #651: Export CSV for course progress not working
The moderator should be able to download the course progress in the CSV file.

adityacp:
This PR closes #651

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -598,7 +598,12 @@ class LearningUnit(models.Model):
     check_prerequisite = models.BooleanField(default=True)
 
     def get_lesson_or_quiz(self):
-        return self.lesson if self.lesson else self.quiz
+        unit = None
+        if self.type == 'lesson':
+            unit = self.lesson
+        else:
+            unit = self.quiz
+        return unit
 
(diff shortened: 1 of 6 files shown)


yaksh/models.py
```diff
@@ -648,6 +648,9 @@ def _create_unit_copy(self, user):
                 order=self.order, type="lesson", lesson=new_lesson)
         return new_unit
 
+    def __str__(self):
+        return self.lesson.name if self.lesson else self.quiz.description
```
prathamesh920: Though not a problem, but in such method it is always better to avoid Foreign key null true. Lesson/Quiz can be null. Or else the case should be handled.
Just a suggestion.

prathamesh920:
Code looks good to me

yaksh/models.py
adityacp: I changed the condition to check for the `type` field in the learning unit model instead of directly checking the foreign key.

prathamesh920: Looks good👍thanks

PR #663, opened 2020-03-16
Add yaml upload support

ankitjavalkar:


yaksh/views.py
```diff
@@ -3155,3 +3155,18 @@ def get_course_modules(request, course_id):
     modules = course.get_learning_modules()
     context = {"modules": modules, "is_modules": True, "course": course}
     return my_render_to_response(request, 'yaksh/course_detail.html', context)
+
+@login_required
+@email_verified
+def download_course_yaml(request, course_id):
+    course = Course.objects.get(id=course_id)
```
adityacp: Please add the condition to check if the user is moderator and course creator or teacher

yaksh/models.py
```diff
@@ -1108,6 +1133,112 @@ def has_lessons(self):
...
+            mod_subitems = []
+            for module in self.get_learning_modules():
+                mod_fields = [
+                    'name', 'description', 'check_prerequisite',
+                    'check_prerequisite', 'check_prerequisite_passes',
+                    'html_data', 'active'
+                ] 
+
+                unit_subitems = []
+                for unit in self.get_learning_units():
```
adityacp: Shouldn't this be `module.get_learning_units()`?

yaksh/views.py
ankitjavalkar: fixed in subsequent commits

yaksh/models.py
ankitjavalkar: Using the course instance method with the same name.

yaksh/models.py
```diff
@@ -232,6 +233,30 @@ def render_template(template_path, data=None):
     return render
 
 
+def _create_yaml(obj, fields, fname, zip_file, obj_type, subitems=None):
+    file_name = '{0}.yaml'.format(fname)
+    course_dict = model_to_dict(obj,
+        fields= fields
+    )
+    relevant_dict = CommentedMap()
```
prathamesh920: Suggestion:
variable name as ```relevant_data```

yaksh/models.py
```diff
@@ -1129,6 +1154,113 @@ def has_lessons(self):
...
 
+    @classmethod
+    def _create_object_from_dict(cls, model_obj, model_type):
+        model_class = get_model_class(model_type)
+        new_obj, obj_create_status = model_class.objects.get_or_create(**model_obj)
+        return new_obj, obj_create_status
+
+    @classmethod
+    def load_yaml(cls, filepath, user):
+        file_list = {}
```
prathamesh920: ```file_list``` is initialized as dict
So variable name can simply be ```files```

yaksh/views.py
```diff
@@ -3237,3 +3237,51 @@ def download_course_progress(request, course_id):
     for student in stud_details:
         writer.writerow(student)
     return response
+
+
+@login_required
+@email_verified
+def download_course_yaml(request, course_id):
+    user = request.user
+    course = Course.objects.get(id=course_id)
```
prathamesh920: ```course_id``` can be invalid then it make break.
Better to use ```get_object_or_404```

yaksh/views.py
```diff
@@ -3237,3 +3237,51 @@ def download_course_progress(request, course_id):
...
+    if not is_moderator(user):
+        raise Http404('You are not allowed to download this course !')
+    if not course.is_creator(user):
+        raise Http404('This course does not belong to you')
+
+    zip_file_path = course.create_yaml()
+
+    response = HttpResponse(content_type='application/zip')
+    response['Content-Disposition'] = dedent(
+        '''attachment; filename={0}_yaml.zip'''.format(course.name)
```
prathamesh920: filename is the course name
course name may contain spaces, better to replace them by underscores

yaksh/views.py
```diff
@@ -3237,3 +3237,51 @@ def download_course_progress(request, course_id):
...
+        response.write(zip_file.read())
+    return response
+
+
+@login_required
+@email_verified
+def upload_course_yaml(request):
+    user = request.user
+    if not is_moderator(user):
+        raise Http404('You are not allowed to download this course !')
```
prathamesh920: The message should ```... not allowed to upload ...```

yaksh/views.py
```diff
@@ -3237,3 +3237,51 @@ def download_course_progress(request, course_id):
...
+
+    if request.method == "POST":
+        form = UploadFileForm(request.POST, request.FILES)
+        if form.is_valid():
+            course_zip_file = request.FILES['file']
+            file_extension = course_zip_file.name.split('.')[-1]
+            if file_extension == "zip":
+                Course.load_yaml(course_zip_file, user)
```
prathamesh920: I think we should have tests for this with different cases.  Since load yaml method may raise an exception if zip is not proper. At least we can handle the known exceptions if any.

yaksh/models.py
```diff
@@ -1129,6 +1154,113 @@ def has_lessons(self):
...
+                if yaml_content.get('type') and yaml_content.get('type') == 'course':
+                    course = yaml_content
+                file_list.update(
+                    {name: yaml_content}
+                )
+        subitems = course.pop('subitems')
+        model_type = course.pop('type')
+        course.update({'creator': user})
+        new_course, course_create_status = cls._create_object_from_dict(course, 'course')
+        new_mod_list = []
```
prathamesh920: name can be simply modules or new_modules or new_mods. No need of _list here.
There are few other places with similar thing.

yaksh/views.py
```diff
@@ -3237,3 +3237,51 @@ def download_course_progress(request, course_id):
...
+                message = "Course uploaded successfully"
+                messages.success(request, message)
+            else:
+                message = "Please upload a ZIP file"
+                messages.info(request, message)
+        else:
+            message = "An error occurred while uploading the file"
+            messages.warning(request, message)
+
+    return my_redirect('exam/manage/courses/')
```
adityacp: Use the `reverse` method instead of manual URL

yaksh/views.py
```diff
@@ -3237,3 +3237,51 @@ def download_course_progress(request, course_id):
...
+        if form.is_valid():
+            course_zip_file = request.FILES['file']
+            file_extension = course_zip_file.name.split('.')[-1]
+            if file_extension == "zip":
+                Course.load_yaml(course_zip_file, user)
+                message = "Course uploaded successfully"
+                messages.success(request, message)
+            else:
+                message = "Please upload a ZIP file"
+                messages.info(request, message)
```
adityacp: This can be a warning message rather than an info

yaksh/views.py
```diff
@@ -3237,3 +3237,51 @@ def download_course_progress(request, course_id):
...
+
+
+@login_required
+@email_verified
+def download_course_yaml(request, course_id):
+    user = request.user
+    course = Course.objects.get(id=course_id)
+    if not is_moderator(user):
+        raise Http404('You are not allowed to download this course !')
+    if not course.is_creator(user):
```
adityacp: Shouldn't the course teacher be allowed to download the course?

ankitjavalkar: Currently they are not, because ownership of uploaded courses is complicating the logic. We'll see the usage of the feature and revisit if required

yaksh/views.py
ankitjavalkar: Done

yaksh/views.py
ankitjavalkar: Done

yaksh/models.py
ankitjavalkar: Done

yaksh/views.py
ankitjavalkar: I have created an issue #668 for this, will send that in as a separate PR

yaksh/views.py
ankitjavalkar: Done

yaksh/views.py
ankitjavalkar: OK, Done using slugify because even though it uses '-' instead of '_' it does a better job of handling special characters and multiple consecutive spaces

yaksh/views.py
ankitjavalkar: Done

yaksh/models.py
ankitjavalkar: Done

yaksh/models.py
ankitjavalkar: Done

adityacp:
After downloading YAML I also got a separate Yaml file for Learning Unit. Is it necessary for uploading as well?

adityacp:
It would be better if a sample template is given so that it will be easy to understand the YAML structure.

adityacp:
Closing this PR. I will reopen it if needed. Substitute feature is being considered i.e. a jupytext based course upload and download.

PR #666, opened 2020-03-18
Add offline support

adityacp:
- Add a new field in lesson model to add relative offline video path
- Add a Vue app to provide an offline interface. Vue app provided from https://github.com/CruiseDevice/offline_yaksh
- Create a zip file to download
- Requires API from the PR #674


yaksh/offline_yaksh/src/components/sidebar.js
```diff
@@ -0,0 +1,74 @@
...
+    if (courseId && quizId && this.gettoken !== undefined){
+      this.fetchQuestions(courseId, quizId)
+    }
+  },
+  methods: {
+    fetchQuestions (course_id, quiz_id) {
+      this.$store.commit('UPDATE_LOADING', true)
+      axios({
+        method: 'get',
+        url: `http://localhost:8000/api/start_quiz/${course_id}/${quiz_id}`,
```
CruiseDevice, line 47: The URL should be of live Yaksh.

CruiseDevice:
I think the offline app in this PR does not have support for int, str, float type questions. I have already created a PR for this in my own repo, but not merged.

PR #667, opened 2020-03-19
Update Yaksh API

CruiseDevice:


api/views.py
```diff
@@ -0,0 +1,413 @@
...
+        if question.type in ['code', 'upload']:
+            json_data = question.consolidate_answer_data(user_answer, user)
+        result = answerpaper.validate_answer(user_answer, question, json_data,
+                                             answer.id)
+
+        # updaTE RESult
+        if question.type not in ['code', 'upload']:
+            print('Non CODE', user_answer)
+            if result.get('success'):
+                print('success')
```
adityacp: Remove print statements

api/views.py
```diff
@@ -0,0 +1,413 @@
...
+        course = Course.objects.get(id=pk)
+        serializer = CourseSerializer(course)
+        return Response(serializer.data)
+    
+
+@api_view(['POST'])
+@authentication_classes(())
+@permission_classes(())
+def login(request):
+    print(request)
```
adityacp: remove print

api/views.py
```diff
@@ -0,0 +1,413 @@
...
+                                             answer.id)
+
+        # updaTE RESult
+        if question.type not in ['code', 'upload']:
+            print('Non CODE', user_answer)
+            if result.get('success'):
+                print('success')
+                answer.correct = True
+                answer.marks = question.points
+                answer.error = json.dumps(result.get('error'))
```
adityacp: Here it should be `answer.error = json.dumps(result.get('result').get('error'))`

PR #671, opened 2020-03-27
Allow stdio expected input to be optional and miscellanous UI fixes

### LINKED ISSUE #658: Addition and fixes in question interface
- Instead of showing the `Add test case` option in the question form, show it below in the test case section.
- Increase textarea height of question description.
- STDIO based formset fails if the expected input field is not filled.

ankitjavalkar:
Closes #658

### DIFF
--- a/yaksh/templates/yaksh/add_question.html
+++ b/yaksh/templates/yaksh/add_question.html
@@ -74,18 +74,6 @@
           </div>
         </td>
       </tr>
-      {% if question %}
-        <tr><td>Add Test Case:</td>
-          <td>
-          <select id="case_type" class="form-control" name="case_type" onchange="frm.submit()">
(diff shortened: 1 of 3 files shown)


yaksh/templates/yaksh/add_question.html
```diff
@@ -127,6 +127,22 @@ <h3>Uploaded files</h3>
         <h3>Test Cases</h3>
       </div>
       <div class="card-body">
+        {% if question %}
```
adityacp: There is a repeated if condition present in the question HTML table. Please remove that.

adityacp:
Also, add `Details&nbsp;<i class="fa fa-angle-down"></i>` instead of just a down icon. It would give better understanding.

PR #674, opened 2020-03-30
Update Yaksh API

CruiseDevice:


api/views.py
```diff
@@ -372,3 +382,42 @@ def delete(self, request, pk, format=None):
         questionpaper = self.get_questionpaper(pk, request.user)
         questionpaper.delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class ICourse(APIView):
```
adityacp: Change the  class name to something else maybe GetCourse

api/views.py
```diff
@@ -184,13 +184,23 @@ def get_answer(self, pk):
...
+        try:
+            if question.type == 'mcq' or question.type == 'mcc':
+                user_answer = request.data['answer']
+            elif question.type == 'integer':
+                user_answer = int(request.data['answer'][0])
+            elif question.type == 'float':
+                user_answer = float(request.data['answer'][0])
+            elif question.type == 'string':
+                user_answer = request.data['answer']
+                print(user_answer)
```
adityacp, line 199: Remove print

PR #676, opened 2020-04-01
Fix custom middleware according to django 3

adityacp:


adityacp:
Merging on a priority basis.

PR #678, opened 2020-04-02
Show prettified R error messages and fix filter

### LINKED ISSUE #679: R clean error messages and filter form
R message should be presented in a better way.
Filter form should take the values of languages and question types from models. Currently the newly added question type or language is not reflected in the filter form.

prathamesh920:
R language available in filter form. Values taken from models.
R messages are prettified like python assertion messages.
Text up to first colon in the R error message removed, as
it is simply a filename or unwanted text.
Closes #679


adityacp:
LGTM

PR #693, opened 2020-04-21
Change UI in grade user

adityacp:
grade_user

- Added a dropdown instead of a complete list
- Added a search bar to search for users


yaksh/templates/yaksh/grade_user.html
```diff
@@ -87,16 +87,17 @@ <h5 data-toggle="tooltip" title="{{course.name}}">
...
-              </li>
-            {% endfor %}
-          </ul>
+        <div id="student" class="col-lg-3">
+          <div class="students-list">
+            <select onchange="window.location.href=this.value" class="custom-select">
+              <option value="">Select student</option>
+              {% for user in users %}
+                 <option value="{% url 'yaksh:grade_user' quiz_id user.user__id course_id %}" {% if user.user__id == data.user.id %} selected {% endif %}>
+                   {{user.user__first_name|upper}} {{user.user__last_name|upper}}
```
CruiseDevice, line 96: If the name is long, it gets mixed with the table on the right-side. This can be ignored cause the name should be considerably longer to break, which is very rare in a real-life case. Other than that looks good to me

prathamesh920:
LGTM

PR #695, opened 2020-04-27
Fix pagination for forum post search

### LINKED ISSUE #694: Issues in Discussion Forum
- If search results in more than 10 threads then pagination will not work for the search results. Clicking on the page number redirects back to the original list.
Maybe create a separate view function for searching posts or add a search request parameter in the paginator.html file.
- Also, add `get_queryset()` before filter method.
- Include paginator before posts list too because the user has to scroll down just for going to the next page.

adityacp:
This PR closes #694


ankitjavalkar:
LGTM

PR #709, opened 2020-05-18
Notifications app for yaksh

CruiseDevice:


notification/tasks.py
```diff
@@ -0,0 +1,62 @@
...
+        if course.is_active_enrollment():
+            message = dedent("""
+                The deadline for the course {0} is {1}, please complete
+                the course if not completed before the deadline.
+                """.format(course.name, course.end_enroll_time)
+            )
+            students = course.students.all()
+            creator = course.creator
+            if students:
+                students_id = students.values_list('id', flat=True)
```
adityacp, line 24: You can directly do values_list as `course.students.values_list('id', flat=True)`

notification/tasks.py
```diff
@@ -0,0 +1,62 @@
...
+                    receiver_ids=students_id, msg_id=nm.id
+                )
+
+
+@task(name='quiz_deadline_task')
+def quiz_deadline_task():
+    courses = Course.objects.filter(active=True, is_trial=False)
+    for course in courses:
+        students = course.students.all()
+        students_id = students.values_list('id', flat=True)
```
adityacp, line 40: Same here

notification/tasks.py
```diff
@@ -0,0 +1,62 @@
...
+        creator = course.creator
+        modules = course.learning_module.all()
+        for module in modules:
+            units = module.learning_unit.all()
+            for unit in units:
+                if unit.type == 'quiz':
```
adityacp, line 46: You don't need to do this much. Just do `course.get_quizzes`

notification/tasks.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+    ), name='course_quiz_deadline_mail_task'
+)
+def course_quiz_deadline_mail_task():
+    courses = Course.objects.filter(active=True, is_trial=False)
+    for course in courses:
+        if course.is_active_enrollment():
+            students = course.students.all()
+            quizzes = course.get_quizzes()
+            for student in students:
```
adityacp, line 96: You don't need to do multiple for loops. Just get all the subscribed students from the subscription table and iterate over that.

PR #713, opened 2020-05-21
Rename celery.py to celery_settings and change imports

adityacp:


adityacp:
Merging on a priority basis. This change is required for deployment.

PR #715, opened 2020-05-21
Check if attempts are allowed and spare time is available before answer is checked

ankitjavalkar:
Changed the function name because - https://www.datacamp.com/community/tutorials/role-underscore-python#SLU

### DIFF
--- a/yaksh/views.py
+++ b/yaksh/views.py
@@ -738,10 +738,17 @@ def check(request, q_id, attempt_num=None, questionpaper_id=None,
         question_paper=questionpaper_id,
         course_id=course_id
     )
+    questionpaper = QuestionPaper.objects.get(id=questionpaper_id)
     current_question = get_object_or_404(Question, pk=q_id)
 
     if request.method == 'POST':
         # Add the answer submitted, regardless of it being correct or not.
+        if paper.time_left() <= 0 or questionpaper.can_attempt_now(user, course_id)[0]:
+            reason = 'Your time is up!'
+            return complete(
+                request, reason, paper.attempt_number, paper.question_paper.id,
+                course_id, module_id=module_id
+            )
         if current_question.type == 'mcq':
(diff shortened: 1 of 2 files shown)


yaksh/views.py
```diff
@@ -844,6 +851,7 @@ def check(request, q_id, attempt_num=None, questionpaper_id=None,
         if current_question.type in ['code', 'upload']:
             if (paper.time_left() <= 0 and not
                     paper.question_paper.quiz.is_exercise):
+                print(">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>")
```
adityacp, line 854: Remove Print

ankitjavalkar: Fixed

prathamesh920:
Implementing this will prevent checking of answer when the quiz auto submits due to time up.
If the user has selected an answer then it should be checked on submit or auto-submit.

Also, I think ```can_attempt_now``` method is not needed here, as this view is checking the answer for the already existing attempt and not requesting for a new one.

yaksh/views.py
```diff
@@ -738,10 +738,17 @@ def check(request, q_id, attempt_num=None, questionpaper_id=None,
         question_paper=questionpaper_id,
         course_id=course_id
     )
+    questionpaper = QuestionPaper.objects.get(id=questionpaper_id)
```
prathamesh920, line 741: This is not needed

prathamesh920, line 741: This is not needed

prathamesh920:
LGTM

CruiseDevice:
LGTM. Coverage seems to be decreasing, pep8 issues and also build fails. Please fix it.

ankitjavalkar:
Finally passes tests after 5 odd rebuilds on travis :(

PR #720, opened 2020-06-01
Creates course through api call

techievivek:
API endpoint to create course.


api/views.py
```diff
@@ -429,4 +431,63 @@ def get(self, request, answerpaper_id, format=None):
...
+                        return Response(unit_serializer.errors)
+                module_instance.learning_unit.set(units)
+            else:
+                return Response(module_serializer.errors)
+        if serializer.is_valid():
+            new_course = serializer.save()
+            new_course.learning_module.set(modules)
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        else:
+            return Response(serializer.errors)
```
adityacp, line 493: Send a 400 status code if there errors in serializers. 400 is for bad requests.

online_test/settings.py
```diff
@@ -215,19 +215,19 @@
...
+#     'DEFAULT_AUTHENTICATION_CLASSES': [
+#         'rest_framework.authentication.TokenAuthentication',
+#         'rest_framework.authentication.BasicAuthentication',
+#         'rest_framework.authentication.SessionAuthentication',
+#     ],
+#     'DEFAULT_PERMISSION_CLASSES': [
+#         'rest_framework.permissions.IsAuthenticated'
+#     ],
+#     'TEST_REQUEST_DEFAULT_FORMAT': 'json'
+# }
```
adityacp, line 230: Don't comment this.

yaksh/migrations/0018_comment_forumbase_post.py
```diff
@@ -0,0 +1,56 @@
+# Generated by Django 3.0.3 on 2020-05-02 08:35
```
adityacp, line 1: Remove all the migration files. They are not needed.

api/views.py
```diff
@@ -429,4 +431,63 @@ def get(self, request, answerpaper_id, format=None):
...
+        for learning_module in request.data['learning_module']:
+            learning_module['creator'] = user_id
+            for learning_unit in learning_module["learning_unit"]:
+                if learning_unit['type'] == 'quiz':
+                    learning_unit['quiz']['creator'] = user.id
+                if learning_unit['type'] == 'lesson':
+                    learning_unit['lesson']['creator'] = user.id
+        serializer = CourseSerializer(data=request.data)
+        modules = []
+        for learning_module in request.data['learning_module']:
```
adityacp, line 452: You can do the serializer part in the previous for loop only where you are setting creator, no need of adding another for loop.

.env
```diff
@@ -1,11 +1,11 @@
 # Django settings
 SECRET_KEY=dUmMy_s3cR3t_k3y
-#DB_ENGINE=mysql
-#DB_NAME=yaksh
-#DB_USER=root
-#DB_PASSWORD=root
+DB_ENGINE=mysql
```
adityacp, line 3: Don't rename .samplenv to .env this is not required it will be done in the production.

api/tests.py
```diff
@@ -904,3 +904,38 @@ def add(a,b):
...
+            'name': 'Python Test Api Course',
+            'creator': 'demo',
+            'enrollment': 'Enroll Request',
+            'learning_module': leaningModule
+        }
+        response = self.client.post(reverse('api:course_create'), data)
+        # Then
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertTrue(
+            Course.objects.filter(name='Python Test Api Course').exists())
```
adityacp, line 937: Check if the learning module and quiz also exists or not.

api/tests.py
```diff
@@ -904,3 +904,38 @@ def add(a,b):
...
+        LearningUnit = {'quiz': quiz, 'type': 'quiz', 'order': 1}
+        leaningModule = {
+            'name': 'LM1',
+            'description': 'module one',
+            'creator': 'demo',
+            'learning_unit': LearningUnit
+        }
+        data = {
+            'name': 'Python Test Api Course',
+            'creator': 'demo',
```
adityacp, line 929: why are you sending a `creator` in the post data for the quiz, learning module?
The creator is required only at the course level.

adityacp:
Your test cases are failing in the Travis ci. Fix the test

api/tests.py
techievivek: LearningModule too have a creator field.

adityacp: I know but anyway creator will be added in the views for the learning module and quiz, so you don't need it in the post data.

PR #724, opened 2020-06-06
Tests for api

techievivek:
Added tests for course creation.


api/serializers.py
```diff
@@ -62,9 +73,14 @@ class Meta:
         model = LearningModule
         fields = '__all__'
 
+    def create(self, validated_data):
+        learning_unit = validated_data.pop("learning_unit")
+        new_module = LearningModule(**validated_data)
```
sunilshetye, line 78: Should this be:

new_module = LearningModule(learning_unit)

Otherwise, the first element is always missed out.

api/serializers.py
```diff
@@ -77,3 +93,9 @@ class Meta:
             'grading_system',
             'view_grade',
         )
+
+    def create(self, validated_data):
+        learning_module = validated_data.pop('learning_module')
+        new_course = Course(**validated_data)
```
sunilshetye, line 99: Should this be:

new_course = Course(learning_module)

Otherwise, the first element is always missed out.

techievivek:
@sunilshetye I am attaching learning unit and module in the view section. Please have a look at the view related to the serializer.

PR #725, opened 2020-06-08
Remove 'copy of' string in _create_lesson_copy()

ankitjavalkar:
- Remove Copy Of in Module, Lesson and Quiz names when duplicating courses

- Prevent student users from accessing Questions

### DIFF
--- a/yaksh/models.py
+++ b/yaksh/models.py
@@ -296,7 +296,7 @@ class Lesson(models.Model):
         lesson_files = self.get_files()
         new_lesson = self
         new_lesson.id = None
-        new_lesson.name = "Copy of {0}".format(self.name)
+        new_lesson.name = self.name
         new_lesson.creator = user
         new_lesson.save()
         for _file in lesson_files:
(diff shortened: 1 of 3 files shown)


CruiseDevice:
@ankitjavalkar can you create the issue for this?

yaksh/models.py
```diff
@@ -296,7 +296,7 @@ def _create_lesson_copy(self, user):
         lesson_files = self.get_files()
         new_lesson = self
         new_lesson.id = None
-        new_lesson.name = "Copy of {0}".format(self.name)
+        new_lesson.name = self.name
```
prathamesh920, line 299: If the name will be same as original lesson name then no need of assigning the same name.

yaksh/models.py
```diff
@@ -575,7 +575,7 @@ def _create_quiz_copy(self, user):
         question_papers = self.questionpaper_set.all()
         new_quiz = self
         new_quiz.id = None
-        new_quiz.description = "Copy of {0}".format(self.description)
+        new_quiz.description = self.description
```
prathamesh920, line 578: Same here, not required

yaksh/models.py
```diff
@@ -932,7 +932,7 @@ def create_duplicate_course(self, user):
         copy_course_name = "Copy Of {0}".format(self.name)
         new_course = self._create_duplicate_instance(user, copy_course_name)
         for module in learning_modules:
-            copy_module_name = "Copy of {0}".format(module.name)
+            copy_module_name = module.name
```
prathamesh920, line 935: Same here

ankitjavalkar:
Changes based on suggestions have been made

PR #737, opened 2020-08-09
[FIX] #546 - create_moderator command if User has no Profile

### LINKED ISSUE #546: create_moderator command fails if User has no Profile
When starting a new project, trying to add a superuser (that has been created using the ```createsuperuser``` command) as a moderator throws the following exception.

```
django.db.models.fields.related_descriptors.RelatedObjectDoesNotExist: User has no profile.
```

Steps to replicate:
- Clone the repository
- Run ```python manage.py migrate``` command
- Run ```python manage.py createsuperuser``` command and create an admin user
- Run ```python manage.py create_moderator admin```

jramnai:
This fixes #546
 
Previously create_moderator command was failing as per the mentioned git issue. So, this PR changes will check for the user profile first, if the user has a profile then it will continue the default flow but if not then will create the profile first(with verified email) and then continue the default flow.

One more change that has been done in the mode query representation in the Django Admin panel. Previously it was fetching the user full name. Now, it will display a full name or username in the Profile model.

### DIFF
--- a/yaksh/management/commands/create_moderator.py
+++ b/yaksh/management/commands/create_moderator.py
@@ -10,6 +10,7 @@ from django.contrib.auth.models import User, Group, Permission
 
 # local imports
 from yaksh.models import create_group
+from yaksh.views import _create_or_update_profile
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


jramnai:
Waiting for the team member to review this.

PR #744, opened 2020-08-26
Avoid duplicate user entry with same email address during upload.

### LINKED ISSUE #743: Upload users csv fix
New user is created with the same email id, if the user had registered previously.
Instead, the old user should be updated.(No new user should be created.)

prathamesh920:
Django allows multiple usernames with same email id.
Preventing this, as we identify users with their email id or username.
closes #743

### DIFF
--- a/yaksh/views.py
+++ b/yaksh/views.py
@@ -2409,8 +2409,10 @@ def _read_user_csv(request, reader, course):
             messages.info(request, "{0} -- Missing Values".format(counter))
             continue
         users = User.objects.filter(username=username)
+        if not users.exists():
+            users = User.objects.filter(email=email)
         if users.exists():
-            user = users[0]
+            user = users.last()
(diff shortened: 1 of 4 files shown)


yaksh/views.py
```diff
@@ -2409,8 +2409,10 @@ def _read_user_csv(request, reader, course):
             messages.info(request, "{0} -- Missing Values".format(counter))
             continue
         users = User.objects.filter(username=username)
+        if not users.exists():
+            users = User.objects.filter(email=email)
```
adityacp, line 2413: Can we use the previous `users` variable here?

prathamesh920: No, because the users is empty then I check for the users with the required email.
Using one filter then it has to be OR condition for the filter and not AND condition.
So to avoid complexity, kept it very simple and easy to understand.

PR #748, opened 2020-08-28
Multiple changes

adityacp:
- Fix a bug where user cannot submit zero as answer
- Fix UI in question statistics
- Fix a bug where the trial question paper was not updated


adityacp:
Merging on priority. Multiple important fixes.

PR #760, opened 2020-09-13
Fix rendering of the quiz list

prathamesh920:


ankitjavalkar:
```test_regrade_by_moderator``` This test is failing, this is not a usual failure. Should I still merge, or can you quickly have a look, in case it does not break something major, I can merge it ASAP

PR #761, opened 2020-09-14
Issuing Digital Certificate

WISE-ELITE-2K18:
## Contributors List
Here is the list of the curators who have designed the Certificate and sent it to User's ground.

* Hima Bhavya Sri Kuchipudi
* Navya Sri Jasti
* Pragna Kanchana
* Priyanka Meesala
* Sai Sruthi Tadikamalla
* Sai Surya Sanjana Addanki
* Sushma Kovvuri
* Tejaswi Tadikonda
* Satya Venkata Dhakshayani Garapati
* Satya Vijaya Sri Thavva
* Vineesha Surya Devi Namana

# Contribution

### Issue Introduction

As the name suggests, a certificate should be sent to the user through digital media as a token of completion of the course. Scholars, you can think "We already have that feature in YAKSH Platform" but this is different from that. We have jotted it down. You people get it, once if you skim through this.

### Why we have chosen this issue?
----------------------------------
These days peeps are expecting the result even without completing their work. But we are not getting the result at the time and as googly we are asked to refill the details but we lazy programmers don't want to do it. We decided to fix this issue, We will be issuing a certificate so that the user can get the certificate the very next minute of the completion of the course. So to send the certificate we may have various ways for all those ways we need to get extra data or information but to make it user friendly or student-friendly we are not taking any bit of extra data. But the people who are reading this, a doubt might have already raised in your mind that how can the user get their certificate. Here is your answer we will be sending the certificate through the mail which the user gives to register into our platform.

### How we fixed it??
-----------------

**1.Certificate Generation**

We have decided to generate the certificate using LaTeX, we can generate it through HTML but to transcend we have used LaTeX.Before embarking into it, the initial step is the LaTeX installation.

Command to install LaTeX:

  ` sudo apt-get install texlive`

   
Certificate:
        
Following is the sample Certificate generated using Latex.

         

**2.Integrating certificate with code**
	 
As the project is totally related to Django and Python the next step is all about how we integrated our TeX file(which we have used to generate certificate) and Python file. As every certificate contains the user name and course name to render those we have sent them as arguments using the Jinja template engine. We have used the jinja2 module to achieve this.And then we have coverted this file to pdf format using pdflatex.

	Command to install Jinja2:
	   pip install jinja2

	Command to install pdflatex:
	   pip install pdflatex

**3.Fetching Data from Yaksh Database:**
	
The jinja2 module is just a means of transport to reach the destination but the source is Yaksh database. Course id and user id are taken from **yaksh_coursestatus** and their corresponding username and course name are fetched from **auth_user** and **yaksh_course** tables.

**4.Sending Certificate via Mail:** 

Once the user is done with the entire course, a certificate has to be sent to his/her mail. And after sending a certificate to the user, the information is updated in the Certificate column in the Course_status table.

**Sending Email:**

Email ids of users are fetched from auth_user using user_id.Finally, An email is sent to user's mail with a message of course completion and with an attachment of a certificate.

	
### Code Modifications
-----------------------
**1.In online_test/online_test/settings.py/**

In this file we have changed the values of variables as:
	 
EMAIL_USE_TLS = True

EMAIL_HOST = 'smtp.gmail.com'

EMAIL_PORT = '587'

EMAIL_HOST_USER = 'yakshcertificate@gmail.com'

EMAIL_HOST_PASSWORD = 'elite2019'

EMAIL_USE_SSL = False

**2. In online_test/yaksh/send_emails.py/**
	
Inorder to send mail to the user, in this file we mentioned the content of the body of email, subject and attachment(of Certificate) which has to be sent.

**3. In online_test/yaksh/models.py/**
	
In this file in the class CourseStatus, we have added a single line of code inorder to add a column to the Course_status table in YAKSH Database.

And in the same class we have created two functions named get_certificateStatus() and set_certificateStatus() to know if the user received his/her certificate or not.

**4.In online_test/yaksh/views.py/**
Here, we imported mail_certificate from send_emails. 

In Views Module, we have included a piece of code to generate a Certificate with his/her details upon Completion of a Course.

**5.In online_test/yaksh/templates/yaksh/course_modules.html**
	
In this HTML file, we have included code to provide a button to get Certificate.


yaksh/models.py
```diff
@@ -1200,6 +1201,14 @@ def set_grade(self):
             grade = grading_system.get_grade(self.percentage)
             self.grade = grade
             self.save()
+    
+    def get_certificateStatus(self):
+        return self.certificate
+
+    def set_certificateStatus(self):
+        self.certificate = True
```
adityacp, line 1209: As you are setting a Boolean value to the certificate field, You should use `models.BooleanField`.

yaksh/templates/yaksh/course_modules.html
```diff
@@ -7,14 +7,14 @@
...
+            <a href="{% url 'yaksh:course_forum' course.id %}" class="btn btn-info pull-right">Discussion Forum</a>
         </div>
         <div class="card-body">
-            {% if course.view_grade %}
+            
                 <center>
-                    <b>Grade:</b> {% if grade %} {{ grade }} {% else %} Will be available once the course is complete {% endif %}
+                    <b>Grade:</b> {% if grade %} {{grade}} {% else %} Will be available once the course is complete {% endif %}
                 </center>
-            {% endif %}
```
adityacp, line 17: I guess you don't have to remove this if condition, because the course creator has the control to enable and disable to view the grades for the course.

yaksh/views.py
```diff
@@ -3106,9 +3127,30 @@ def view_module(request, module_id, course_id, msg=None):
     context['course'] = course
     context['state'] = "module"
     context['msg'] = msg
+    course_status = CourseStatus.objects.filter(
+            course_id=course.id, user_id=user.id)
+    print(user.get_full_name(), course , user.email)
+    #print(course_status.first().get_certificateStatus())
```
adityacp, line 3133: Remove unnecessary print statements.

yaksh/views.py
```diff
@@ -3106,9 +3127,30 @@ def view_module(request, module_id, course_id, msg=None):
...
+    if course.percent_completed(user, all_modules) == 100 and course_status.first().get_certificateStatus() == '':
```
adityacp, line 3134: `get_certificateStatus()` should give a boolean value and not a string. If you set the certificate to the BooleanField that will be solved.

yaksh/views.py
```diff
@@ -3106,9 +3127,30 @@ def view_module(request, module_id, course_id, msg=None):
...
+        return redirect('/')
+    if course.percent_completed(user,all_modules) == 100 and course_status.first().get_certificateStatus() == 'True':
+        messages.warning(request, 'You have already recieved your certificate, kindly check your mail')
+        return redirect('/')
     return my_render_to_response(request, 'yaksh/show_video.html', context)
 
 
+
+
+
```
adityacp, line 3153: As per PEP8 standards, only two new lines between functions are required.

yaksh/views.py
```diff
@@ -3580,3 +3622,24 @@ def extend_time(request, paper_id):
     messages.info(request, msg)
     return my_redirect('/exam/manage/monitor/{0}/{1}/'.format(
         anspaper.question_paper.quiz.id, course.id))
+
```
adityacp, line 3625: I guess you have added lot of extra lines. Please remove it

.env
```diff
@@ -0,0 +1,13 @@
...
+DB_NAME=yaksh
+DB_USER=yaksh_user
+DB_PASSWORD=mysecretpassword
+#DB_HOST=yaksh-db
+DB_PORT=3306
+# Yaksh settings
+N_CODE_SERVERS=5
+SERVER_POOL_PORT=53579
+SERVER_HOST_NAME=http://yaksh-codeserver
+#SERVER_TIMEOUT=4
```
adityacp, line 13: Don't change  the `.sampleenv` to `.env`. This will lead to issue while setting up development environment, because mysql may not be used in the dev env.

online_test/settings.py
```diff
@@ -122,20 +122,25 @@
...
-EMAIL_HOST = 'your_email_host'
+EMAIL_PORT = '587'
 
-EMAIL_PORT = 'your_email_port'
+EMAIL_HOST_USER = 'yakshcertificate@gmail.com'
 
-EMAIL_HOST_USER = 'email_host_user'
+EMAIL_HOST_PASSWORD = 'elite2019'
 
-EMAIL_HOST_PASSWORD = 'email_host_password'
```
adityacp, line 134: Please remove  the Host user and Host password. Username and Passwords should never be released in public. Let this be as it is.

online_test/settings.py
```diff
@@ -122,20 +122,25 @@
...
+EMAIL_USE_SSL = False
 
 # Set EMAIL_BACKEND to 'django.core.mail.backends.smtp.EmailBackend'
 # in production
-EMAIL_BACKEND = 'django.core.mail.backends.dummy.EmailBackend'
+EMAIL_BACKEND = 'django.core.mail.backends.smtp.EmailBackend'
```
adityacp, line 143: Keeping this to smtp will require a email backend even in dev env. So keep it as dummy. On production server this will be changed.

adityacp:
There might be an issue while generating the pdf file. I found out that the file name is certificate.pdf. Can you please check that no two persons gets the same certificate. I think the certificate.pdf file will be overridden every time. It would be good if the certificate  name is generated dynamically i.e. it can be `certificate_<date_time>` or something else which can be unique in nature.
Also, you can check if the certificates pdf files can be in memory files. The certificates are now created on the disk which will be a problem if there are a lot of certificates. If in memory files are not possible then you can add a specific directory for certificates.

online_test/settings.py
WISE-ELITE-2K18: Shall we remove email_port and email_host

WISE-ELITE-2K18:
> * There might be an issue while generating the pdf file. I found out that the file name is certificate.pdf. Can you please check that no two persons gets the same certificate. I think the certificate.pdf file will be overridden every time. It would be good if the certificate  name is generated dynamically i.e. it can be `certificate_<date_time>` or something else which can be unique in nature.
> * Also, you can check if the certificates pdf files can be in memory files. The certificates are now created on the disk which will be a problem if there are a lot of certificates. If in memory files are not possible then you can add a specific directory for certificates.
> * There are no test cases for the feature. Please add a few test cases.

yaksh/models.py
```diff
@@ -1201,6 +1202,13 @@ def set_grade(self):
             self.grade = grade
             self.save()
 
+    def get_certificateStatus(self):
+        return bool(self.certificate)
```
CruiseDevice, line 1206: No need to convert it to bool. Certificate is already a BooleanField.

PR #775, opened 2020-10-07
Set prereqs to False by default

ankitjavalkar:


prabhuramachandran:
@ankitjavalkar -- can you also show this on the UI so it is changeable easily?

PR #797, opened 2020-11-12
Fix video issues

adityacp:
- Fix issue onclick toc jump to player time.
- Use bootstrap modal instead of jquery dialog for responsive interface.
- Disallow empty question submission for lesson quiz.


adityacp:
Merging on priority.

prabhuramachandran:
No need to make it compulsory to answer, just that if they don't then the next time it should pop up again.

PR #816, opened 2021-02-10
UI and Bug Fixes

adityacp:
- Revamp UI in course page.
- Fix a bug where a trial module is created without setting is_trial in models.
- Fix a bug in question statistics where only completed answerpaper data is needed.
- Fix a error in monitor when dataframe does not have the column


adityacp:
Merging on Priority.

PR #828, opened 2021-03-24
Fix in QR code generation

adityacp:
Make QR code generation compatible with local storage or S3 storage.


adityacp:
Merging on priority

Issue #835, opened 2021-04-12
Exception when select question in "Design Question Paper" page


thaihaquang:
I faced some issues when I was trying to select question. Below is the details
![image](https://user-images.githubusercontent.com/42860711/114361859-decd4800-9ba0-11eb-9dbe-41d56804adfa.png)

I checked the source code and I could see some problems
![image](https://user-images.githubusercontent.com/42860711/114361894-ebea3700-9ba0-11eb-85d4-565435d23d45.png)
The questions may be undefined if the question_type is none.
Could you please help me to fix it? 
Thanks !

adityacp:
Can you tell us what action triggered this? Maybe a small GIF would help.

PR #837, opened 2021-04-15
Fix files issue

adityacp:
- Fix issue for deleting and saving lesson and question files with S3.
- Remove unnecessary model methods.
- Fix issue #835


adityacp:
Merging on priority.

PR #839, opened 2021-04-21
Allow creator or teacher access to course without enrolment

ankitjavalkar:


yaksh/views.py
```diff
@@ -3185,7 +3185,9 @@ def view_module(request, module_id, course_id, msg=None):
 def course_modules(request, course_id, msg=None):
     user = request.user
     course = Course.objects.get(id=course_id)
-    if user not in course.students.all():
+    user_is_not_student = user not in course.students.all()
```
adityacp, line 3188: Use `course.is_student()` mehtod.

